PR #4, opened 2015-09-24
Composer autoloader

fiammybe:
Changes to start using the composer autoloader. This doesn't work on its own (I have an inclusion of the composer php library planned in the future), so a composer client is needed.
The composer autoloader will need to be extended for the inclusion of all the libraries that we use (only changing the libraries folder will not be enough, there are some includes that will need to go).


skenow:
There are conflicts, @fiammybe - I don't know if they were there before I merged in the controller feature from @MekDrop.

skenow:
The changes to the adsense files and their references to analytics has me wondering if the adsense feature is being changed.

fiammybe:
you are right, this is one of those merge requests where several tickets have been mixed, I thought I sorted it out, but it seems not to be the case. I'll close this merge request, and start over from the current state of the retro branch (with the last merges).

PR #5, opened 2015-09-24
Analytics ga unified

fiammybe:
update of the google analytics script to the unified or universal version. More flexible handling of these kinds of scripts can be done later.


fiammybe:
A better pull request (cleaner) is forthcoming

PR #8, opened 2015-09-26
Integrate the new Admin theme by Debianus

fiammybe:
Debianus did a wonderful job integrating a new admin theme for 2.0, let's include it in the alphas!


skenow:
@fiammybe Do we want to make sure the HTML and CSS validate before or after merging?

fiammybe:
I would do the merge, to get what we have already in the retro branch. Any
improvements will be bug tickets. If we keep it apart until it is deemed
perfect, it will have taken so long that it will be unmergeable.

Op di 6 okt. 2015 01:58 schreef Steve <notifications@github.com>:

MekDrop:
+1 Let's merge! I think this question about having bugs must be only for core changes, otherwise we can always create new tickets :)

skenow:
There does need to be some criteria for this. Otherwise, why fork, create
branches, and submit a pull request? Just commit to the trunk and be done
with it.

MekDrop:
I mean for new commits that can have only some minor issues (like template compatibility with standards), we could merge faster without spending much time for testing. I think it's better to move forward faster because it's more than 6 years we are developing ICMS 2.0... I think we need somehow release it.:)

fiammybe:
You're right @MekDrop we need to up the speed on this, but @skenow is also right that we need to have some procedure on how to organise development. I invite you both to the forums (http://community.impresscms.org/modules/newbb/viewtopic.php?post_id=49958#forumpost49958) where we can discuss this in more detail, and not in the margin as discussions on a merge request.

PR #14, opened 2015-10-15
Added @property doctype style comments in icms library for better autocompletion of virtual properties

MekDrop:
ref #885


MekDrop:
uff... I think I have replaced bad tags to correct ones everywhere in ICMS library.

MekDrop:
@skenow, @fiammybe would be ok if I would merge this one? This pull request also fixes some bad line endings for some files and that's why I'm getting some issues working on other things (there are some possibilities for getting many merge conflicts if I would start to do something without merging this pull request first)

MekDrop:
Thank you!

PR #15, opened 2015-10-16
Added virtual properties for system module objects

MekDrop:
For ticket #886 

This just adds @property lines for documentation generation and better auto-completion support. Also it removes some duplicated other lines with @ from class description


MekDrop:
Comments for this fixed too!

PR #17, opened 2015-10-20
compatibility for `XoopsDatabaseFactory::getDatabaseConnection`

nao-pon:


nao-pon:
@fiammybe Thanks! :+1:

PR #18, opened 2015-10-20
compatibility for non PDO `XOOPS_DB_TYPE`

nao-pon:


fiammybe:
Hi @nao-pon , many thanks for your proposed changes. Can you explain why this change is needed? I see what you change, but I do not understand why.

Thank you very much.

nao-pon:
Hi @fiammybe :laughing:

I caught a error `class 'SqlUtility' dose not exists` on module install of my develop module X-elFinder. (My ICM `XOOPS_DB_TYPE` is 'mysql')

That code is

```php
include_once XOOPS_ROOT_PATH.'/class/database/sqlutility.php' ;
$sqlutil = new SqlUtility ;
```

full code is https://github.com/nao-pon/xelfinder/blob/master/xoops_trust_path/modules/xelfinder/oninstall.php

Thanks for your support.

nao-pon:
I changed code.

skenow:
The proposed change appears OK to me - it covers when the calling code has not defined $driver as it should have. Remember - this is a legacy portion of the code and it is dropped in the 2.0 branch.

nao-pon:
@fiammybe , @skenow Thanks! :+1:

PR #19, opened 2015-11-15
#887 php unit tests for icms library

MekDrop:
I hope all these tests will be improved in the future because not everything here are tested. In most cases I these my tests verifies only that classes, variables, constants exists. I think everything it's possible to write and better way... but I believe this is enought for initial testing nad test writing.

Some tests will fail. Because there are some bugs in our libraries (in most cases bad variable types definitions). Because some of these are easy to fix, I'll try to post separate merge requests.

Because fatal errors in libraries/icms/images/ classes right now is impossible to run all tests at once. However it's possible to run every test separately with such command line (from root tests folder):
`phpunit -c configuration.xml libraries/icms/FILE.php`


fiammybe:
I tried to merge this on a clean retro branch, and merging was ok. The site still installs after merging. I didn't get my test environment up (PHPUnit 4.8 needs PHP 5.6, which I currently don't have) but I believe we can merge this one in, if only to have a preliminary test framework already available. Bugs in the tests can be done on a case-by-case basis.
@skenow I let you decide if we need more verifying or if this can be merged as-is into retro.

skenow:
The changes introduced by this are additions, separate from the functionality of the core - I don't see this interfering with normal operations.

PR #23, opened 2015-12-02
 #894 - Rewrite errors.php with new icms response classes

MekDrop:
Errors.php now is rewritten with new icms_response functionality.

Also, these chances were also made:
 * Fixed bug with custom 404 error
 * `xoops_header` and `xoops_footer` now are deprecached, use also icms_response functionality and system_blank.html template 
 * `\icms::$response` has instance of main response
 * `icms_cp_header` and `icms_cp_footer` also were rewritten to use icms response classes
 * `header.php` and `footer.php` also rewritten
 * removed few warnings generated by prefixing variables, constants with @ symbol 
 * rewrtitten admin menu generator function to run a bit faster and to be easier readable
 * few old comments replaced and few others removed

*Please test with current modules!*


skenow:
This is a very confusing set of commits - I'm not sure how to start to evaluate it. Even github shows a lot of 'empty files modified' that are essential files, and Scrutinizer shows 36 new issues. I guess I'll start by looking at each commit and commenting there.

MekDrop:
@skenow about naming convention: that works right now fine ;) I don't mind to rename in old way, but from otherside in near future we will need to rename it with namespace because composer PSR-0 standard was deprecated and PSR-4 doesn't support such conventions.

For PSR-4 is required to define classes in such way:
`\<NamespaceName>(\<SubNamespaceNames>)*\<ClassName>`

So, basically `icms_response_Defaultemptypage` should be renamed into `\ImpressCMS\Response\DefaultEmptyPage` or something similar name.

MekDrop:
@skenow, yeah kind of. You can ignore all these empty files. I think these files marked as changed because some configuration issues with vagrant (I'm using it)... I'm not sure why this happens... Maybe you could test if this merge request works with your favourite modules... or just check files that could be linked to the change log at the start :|

MekDrop:
After changing line-ends it seems this merge request became easier to read ^_^

skenow:
Installed imTagging, imBlogging, and SimplyWiki - they all installed successfully and the issue I had with the install function not adding the initial page and page profiles for SimplyWiki has been fixed.

However, there are some fatal errors --
Fatal error: Call to a member function addScript() on a non-object in /var/www/mekdroperror/include/calendarjs.php on line 52. The missing object is $icmsTheme and it occurs when trying to add a new post on both the admin side and user side of imBlogging.

MekDrop:
Thanks, I totally forgot to check calendar functionality :| I'll will look into

MekDrop:
It seems that bug was because there were missing bindings with $icmsTheme. Now it works. I have verified with *imblogging* and *news_ipf* modules

PR #24, opened 2015-12-02
#120 inclusion of .htaccess files that block php execution in appropriate directories

MekDrop:
Almost the same code that was in description of [this ticket](https://www.assembla.com/spaces/impresscms/tickets/cardwall#ticket:120) added to .htaccess. 

I decided to not add hot linking protection as was described there, because not everyone needs. But other parts exists. Also direct access to templates_c/*, theme/*.php is now prevented.


fiammybe:
Cool. This one definitely comes in the 'security' category. Would there be
similar things for nginx and IIS?

Op wo 2 dec. 2015 23:21 schreef Raimondas Rimkevičius <
notifications@github.com>:

MekDrop:
Yes, but I don't think so that we can supply such configuration for these platforms out of box. I think these servers doesn't have possibility to include supplied configuration on the fly.

From other side, I think users of these servers could use tools similar to this one -> http://winginx.com/en/htaccess to convert .htaccess to needed configuration

skenow:
Does this close ticket 120? The commit messages can be used to update tickets so we don't have to update the tickets separately

MekDrop:
@skenow yes. I forgot how to use that functionality so I'm doing that with my hands.

PR #28, opened 2015-12-03
#896 - Move config category constants to class

MekDrop:
ICMS_CONF_* -> \icms_config_Handler::CATEGORY_*


skenow:
I'm working on the conflict now - it is in plugins/preloads/autotasks.php. I'll test locally, then update

PR #30, opened 2015-12-04
Imported config to .scrutinizer.yml 

MekDrop:


fiammybe:
Cool, I talk about it, and before I have the opportunity to look into it, you already have it done. Thanks a lot.

MekDrop:
NP ;) 

It almost same as http://wiki.impresscms.org/modules/wiki/index.php?page=Coding_Standards except here is no "Maximum Line Length" (I have no idea how specify), and I decided to ban tabs and use spaces instead, because this answer -> http://programmers.stackexchange.com/questions/57/tabs-versus-spaces-what-is-the-proper-indentation-character-for-everything-in-e 

4 spaces is new indentation standard for us like in PSR, YAML and most other places. I think this will prevent all developers to forget where what spaces should be.

MekDrop:
Thanks @skenow for the link to very usefull page in our Wiki.

skenow:
I know we've already added this, but here's a question, now that I see the options - for all other constants, we use UPPERCASE. Why not for TRUE, FALSE, and NULL?

MekDrop:
Main problem is that there are nothing about such possibility in scrutinizer  documentation. I think the only way to do that is to write own PHPCS configuration file.

From otherside I think I'm addicted to the lower-case form. Maybe, it's because of [this old reason](http://stackoverflow.com/a/3807178/1762839). Maybe it's because on stackoverflow.com answers people usually use lowercase. Maybe it's because of Yii, Symfony 2, CakePHP (I have used these frameworks). Maybe, it's because JS and JSON says to use lowercase too. So, I have contributed everything in lowercase. Sorry for that. But if you still want to apply upper-case as requirement, please give something that could help me and others like me to change such writing behaviour :|

skenow:
I replied in Slack, and I'll put here for reference - TRUE, FALSE and NULL are not reserved constants, at least not in earlier versions of PHP. So, they can be assigned values other than true, false and null. The uppercase style is used in a majority of places, since that was the standard as we were applying the code standards. When appropriate, switching to lowercase is acceptible.

Something else I noticed in this is the ternary operator options in this file (this is not specified in our code standard, nor is it specified in Zend's). All other operators in our standards and theirs, include a space around all the operators, and I find it more readable with the spaces before and after ? :

PR #41, opened 2015-12-07
#905 DTYPE_ARRAY doesn't convert all values propertly (fix)

MekDrop:


MekDrop:
@skenow, @fiammybe would be ok to merge this? :)

PR #42, opened 2015-12-07
Fix bug with assigning properties for objects

MekDrop:
Fixed that bug + few optimizations and comments


MekDrop:
If I remember correctly with this pull request is one problem: is not solves ticket #893. However it solves some other problems. So... would it be a good idea to rename it?

fiammybe:
@MekDrop yes, please rename it so that it describes better what is included

MekDrop:
Fixed #84 issue

PR #44, opened 2015-12-09
Adding editorconfig Test #888

fiammybe:
Adding editorconfig Test #888

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,8 @@ end_of_line = lf
 insert_final_newline = false
 charset = utf-8
 indent_style = tab
+trim_trailing_whitespace = true
+tab_width = 4
 
 [{package.json,.travis.yml}]
 indent_style = space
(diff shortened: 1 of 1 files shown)


fiammybe:
Depending on the editor, you may still need a plugin to digest the contents of these files. HAve a look at http://www.editorconfig.org

MekDrop:
Awesome! I'll give a try in next few hours.

MekDrop:
It's a bit confusing for NetBeans users (there are no notifications about that editorconfig is enabled for project but it works) right now but it seems it works! ^_^

Few corrections for global configuration:
```
trim_trailing_whitespace = true
tab_width = 4
```

First line will automatically trim tabs or spaces at end of the line.
Second is needed just to make a sure that editor is displaying all tabs as max 4 chars wide (not 2 and not 8). I think this would solve all problems with formatting PHPDoc comments.

PR #45, opened 2015-12-10
feature/863-composer-autoloader

fiammybe:
Replace the homegrown ImpressCMS autoloader with the composer one. No other integration of external libraries as of yet, those will need to be done in different tickets.


MekDrop:
I'm unsure why Travis CI failed on this ticket but I don't see here anything bad. I think we can merge this pull request.

MekDrop:
I have again executed TravisCI test and everything it passed. I believe that this was some internal TravisCI issue.

fiammybe:
Is it ok to merge this then?

htdocs/libraries/icms/Autoloader.php
```diff
@@ -8,6 +8,9 @@
  * @since	1.3
  * @author	Marc-André Lanciault (aka marcan) <mal@inboxintl.com>
  */
+
+include_once ICMS_ROOT_PATH . "/vendor/autoload.php"
+
```
skenow: There is a fatal error at this point, as pointed out by Scrutinizer. There is a missing ; at the end of the line

htdocs/install/page_tablesfill.php
```diff
@@ -1,74 +1,78 @@
...
+$dbm = new db_manager();
+
+if (!$dbm->isConnectable()) {
+	$wizard->redirectToPage( 'dbsettings' );
+	exit();
+}
+
+icms::$db = &$dbm->db;
+icms::$xoopsDB = &$dbm->db;
+
```
skenow: I'm not sure why these are marked by Scrutinizer as issues. The problem it identifies is that ->db is not accessible because it is private, which it is not. It is declared with 'var', which is public.

MekDrop: I think this is Scrutinizer bug :/ 

However another your comment about missing ; is real issue. Somehow I have missed that small but at the time big issue when I wrote my first comment here.

skenow: Good to have more eyes on this

alvinbunk:
Just looking at the files, I'm not certain; but I'm wondering if because on line 54 of dbmanager.php the icms_db_legacy_Factory::getDatabase() function gets a reference copy. Possibly "var $db" needs to be changed to "public var $db" or maybe implement a get_db() function in dbmanager.php?

By the above, I'm just guessing. Maybe someone should try this out.
The other thing to check is if one's IDE is properly setup, does it report any errors with this code. I can't see that Scrutinizer would have this kind of bug :|.

skenow:
When a property is declared as a var, and not explicitly as public,
protected, or private, PHP treats it as public. It is possible Scrutinizer
does not recognize the property as being inherited from the parent class
and gives it a private scope. Not sure.

MekDrop:
Because I'm working on DB part, please skip all changes. Hopefully this issue will be fixed by me automatically in the future ;)

fiammybe:
@mekdrop do you want us to leave this PR on the side for the time being?

MekDrop:
Why? It's hard to fix? Or any other reason? :| 

At least some basic composer integration would help me in near future to achieve easier some planned things.

fiammybe:
@MekDrop I wasn't sure what you meant with "please skip all changes" :-) I'll fix the ';' and let you know so it can be verified and merged.

MekDrop:
@fiammybe, ah, it's was for [this comment](https://github.com/ImpressCMS/impresscms/pull/45#issuecomment-167933485) comment ;)

PR #46, opened 2015-12-14
#904 DTYPE_FILE doesn't work as expected

MekDrop:
Fixes to pass php unit test


MekDrop:
I have updated this branch to solve conflicts. If is OK, we could finally merge and fix the last part of not working tests (however this will not solve issue with the need to run tests on already installed CMS)

MekDrop:
Would it be ok to merge this commit if nobody disagrees?

PR #49, opened 2015-12-29
Add IDs to system_userinfo.html file

alvinbunk:
Pull for my own copy.

### DIFF
--- a/htdocs/modules/system/templates/system_userinfo.html
+++ b/htdocs/modules/system/templates/system_userinfo.html
@@ -6,18 +6,18 @@
   <tr align="center">
     <td>
     <{if $icms_isadmin == true}>
-    <input type="button" value="<{$lang_administration}>" onclick="location='admin.php'" />
+    <input type="button" value="<{$lang_administration}>" onclick="location='admin.php'" id="userinfo_admin" />
     <{/if}>
-    <input type="button" value="<{$lang_editprofile}>" onclick="location='edituser.php'" />
-    <input type="button" value="<{$lang_avatar}>" onclick="location='edituser.php?op=avatarform'" />
-    <input type="button" value="<{$lang_notifications}>" onclick="location='notifications.php'" />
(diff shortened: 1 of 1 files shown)


alvinbunk:
Need this to help with Test Cases Steve.

MekDrop:
Is not too early to merge this pull request? I see here some id's added however i don't see anything how they are helping to do anything :|

alvinbunk:
MekDrop, IDs have just been added. The Selenium (or whatever tool) will use the IDs on the page to run it's tests.

MekDrop:
Yeah, I understand that however I believe that would be a better if you add in this pull request also at least one test were you are using these new IDs. I think it's a better practice than creating pull requests for very small pieces of the code because some parts of your work can change before you do some final strokes. 

Also, I think if you install [profile module](https://github.com/ImpressCMS/impresscms-module-profile) there is a possibility to add for this page some extra fields. I can't remember now how this works, but I think you should look into and verify that your tests are working when this module is installed too.

MekDrop:
Also, please rename this pull request into something meaningful :)

fiammybe:
@alvinbunk could you rename please?

alvinbunk:
I renamed it!

fiammybe:
Can anyone tell me the status of this? I'm very interested in this idea of automatic tests, but this seems to have been hanging here these last months. I'd like to see it happen :-)

MekDrop:
This pull request doesn't do anything very useful however I don't see any reason why we shouldn't merge it.

fiammybe:
Then we can merge it, as far as I'm concerned. @alvinbunk did the work to prepare it, we should be happy that he wants to, and integrate his work unless it poses a problem (which it doesn't) :-)
Op 12/04/2016 9:32:45, Raimondas Rimkevičius <notifications@github.com> schreef:
This pull request doesn't do anything very useful however I don't see any reason why we shouldn't merge it.

alvinbunk:
FYI, I'm not waiting on this; I've just been busy with a school project that will be taking all of my time. So I won't be looking at the Selenium tests for a long time. It's a good thing though that this file will be merged in. There are a lot of PHP frameworks that have the ability to add/specify id or class values for html elements.

PR #50, opened 2016-01-04
#911 selecting new theme causes index.php main page to go blank

MekDrop:
Bug was in theme factory, createInstance function. There was one case when boolean instead of string for theme name was used.

Also my github pull found some bad line endings and unrequired spaces I added these automatically generated chages to this pull request.


fiammybe:
I tested only the changes in the Theme Factory, and those work ok, the problem is solved now. However, I don't think it's a good idea to have all the space fixes and stuff in this pull request as well, as they have nothing to do with the resoluion of the ticket. They don't break anything, but it's not clean and will lead to confusion later.

MekDrop:
ok, I'll create two pull requests for this

alvinbunk:
I tried your changes out on my system @MekDrop and it fixes the selecting Themes issues. One thing I noticed though, when you select the theme from the home page (click on the theme in the list area), the theme changes but you get a "Error 404 Not Found" page error. Just letting you know.

MekDrop:
@alvinbunk see merge request #51 There are less code. And with that branch everything for me works fine :)

PR #60, opened 2016-04-11
ticket 912 - added storing of the registration moment to the adduser functionality in the backend.

fiammybe:
ticket 912 - added storing of the registration moment to the adduser functionality in the backend.


MekDrop:
Just curious: why date('U') but not time() ? I think haven't used date('U') :/

fiammybe:
I was looking for a way to have the epoch, and date('U') gave me that. No other reason.

fiammybe:
Just thinking, would it be better to define a default value in the base user class (icms_Member_user_Object)? that should cover more cases in 1 go

MekDrop:
Yeah that would be a better choice

fiammybe:
I took this in pull request 61

PR #61, opened 2016-04-11
feature/912-member-register-date-1970

fiammybe:
add a default value for the registration date of the user in the base class.

### DIFF
--- a/htdocs/libraries/icms/member/user/Object.php
+++ b/htdocs/libraries/icms/member/user/Object.php
@@ -38,11 +38,11 @@ defined('ICMS_ROOT_PATH') or exit();
 
 /**
  * Class for users
- * 
+ *
  * @author	Kazumi Ono <onokazu@xoops.org>
  * @copyright	Copyright (c) 2000 XOOPS.org
  * @package	ICMS\Member\User
- * 
+ *
  * @property int    $uid               User ID
  * @property string $name              Name
  * @property string $uname             Username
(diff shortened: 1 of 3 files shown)


fiammybe:
Please test

MekDrop:
I think this fix will work. However it's not a good way to fix this issue. Because unnecessary date('U') will be called always also for already created users before data from database is read.

I thinking about somehow to add caching for fields definitions and in that case this will solve current issue with your code. However in that case possible a new issue will rise: date will be cached also.

fiammybe:
I agree that the date is unnecessary when the data is there, but that's what default values are for I believe : you fill the field with the default, and that default is overridden by the data retrieved from the DB or by the user on creation of a new entry.

Can we fix this problem another way without rewriting a part of the core? If the default is not ok, I'll go back to the other fix : add the save to the admin page.
Op 12/04/2016 9:42:50, Raimondas Rimkevičius <notifications@github.com> schreef:
I think this fix will work. However it's not a good way to fix this issue. Because unnecessary date('U') will be called always also for already created users before data from database is read.
I thinking about somehow to add caching for fields definitions and in that case this will solve current issue with your code. However in that case possible a new issue will rise: date will be cached also.

MekDrop:
With one user there are no problems at all, but if you want do something with many users such calls in default values slows down ICMS overall. 

Take a look to this file -> https://github.com/ImpressCMS/impresscms-module-imblogging/blob/master/class/post.php  It uses `beforeSave` event. I think it's the best way to solve custom callable issue to use such way. Also there is another event that can be useful for this `beforeInsert`. Both these events shares same signatures.

But from other side, your change is good enough to merge. 

*Hint:* If you decide to upgrade again, don't close this pull request. Just commit required changes in same branch and write some comment here ;)

fiammybe:
I think I'll look into the beforeSave anyway. 
I generated the pull request from within PHPStorm and I wasn't sure if the changes I made to the files would be visible in the PR after that :-) I'll try now.
Op 12/04/2016 10:50:45, Raimondas Rimkevičius <notifications@github.com> schreef:
With one user there are no problems at all, but if you want do something with many users such calls in default values slows down ICMS overall.
Take a look to this file -> https://github.com/ImpressCMS/impresscms-module-imblogging/blob/master/class/post.php [https://github.com/ImpressCMS/impresscms-module-imblogging/blob/master/class/post.php] It uses beforeSave event. I think it's the best way to solve custom callable issue to use such way. Also there is another event that can be useful for this beforeInsert. Both these events shares same signatures.
But from other side, your change is good enough to merge.
Hint: If you decide to upgrade again, don't close this pull request. Just commit required changes in same branch and write some comment here ;)

htdocs/libraries/icms/member/user/Object.php
```diff
@@ -117,7 +117,7 @@ public function __construct(&$handler, $data = array()) {
         $this->initVar('email', self::DTYPE_STRING, null, true, 60, null, null, _US_EMAIL);
         $this->initVar('url', self::DTYPE_STRING, null, false, 255, null, null, _US_WEBSITE);
         $this->initVar('user_avatar', self::DTYPE_FILE, null, false, 30, null, null, _US_AVATAR);
-        $this->initVar('user_regdate', self::DTYPE_INTEGER, date('U'), false, null, null, null,
+        $this->initVar('user_regdate', self::DTYPE_INTEGER, null, false, null, null, null,
```
MekDrop: add such method to Handler and everything should work:
```
protected function beforeInsert(&$obj) { 
   if (!$obj->user_regdate) {
       $obj->user_regdate; 
   }
}
```

MekDrop: I have updated my previous comment ;) Forgot to write a bit more code there to make it work.

PR #63, opened 2016-05-18
HTMLPurifier using Composer

fiammybe:
Use HTMLPurifier managed with Composer. Updating a new version of HTMLPurifier becomes as easy as 'composer update'


MekDrop:
Works fine! I think we can merge.

PR #72, opened 2016-11-27
Correct link to the documentation page on the wiki now see #930

fiammybe:
Fix #930


skenow:
Would you check the line endings? Every line shows as a change.

fiammybe:
Line endings were set to Windows. I need to see if the Git setting to
automatically convert its set ok, otherwise I'll change the line settings
on my dev environment

Op zo 27 nov. 2016 16:04 schreef Steve <notifications@github.com>:

PR #91, opened 2017-02-17
Fix caching bug

jegelstaff:
If more than one $perms is returned, then only the last one was being cached, therefore subsequent requests would give incorrect results. This change caches the entire set of $perms.


jegelstaff:
This probably affects multiple branches!

jegelstaff:
Cool :)

PR #96, opened 2017-04-04
Use reg users group if no default module/page set

jegelstaff:
Currently, if the Registered Users group has a default module/page set, it is ignored if the user is a member of another group that has no default module/page set.

Perhaps the Registered Users group should be used as a failsafe, so that you don't have to set default pages for every single group in your site? The only downside I can see is that perhaps webmasters would have specifically set a certain group to have no default page and intend that as the behaviour for that group. I think it's more powerful to rely on Registered Users to avoid people having to set a default page for every group though. Especially on sites with many many groups.

### DIFF
--- a/htdocs/index.php
+++ b/htdocs/index.php
@@ -75,12 +75,12 @@ if (preg_match_all('|([^/]+)/([^/]+)/([^/]+)(.*)|', $path, $params, PREG_SET_ORD
     );
 
     // added failover to default startpage for the registered users group -- JULIAN EGELSTAFF Apr 3 2017
-	$groups = (!empty(\icms::$user) && is_object(\icms::$user)) ? \icms::$user->getGroups() : array(XOOPS_GROUP_ANONYMOUS);
+	$groups = (!empty(\icms::$user) && is_object(\icms::$user)) ? \icms::$user->getGroups() : array(ICMS_GROUP_ANONYMOUS);
 	if(($icmsConfig['startpage'][$group] == "" OR $icmsConfig['startpage'][$group] == "--") 
-	AND in_array(XOOPS_GROUP_USERS, $groups) 
-	AND $icmsConfig['startpage'][XOOPS_GROUP_USERS] != "" 
-	AND $icmsConfig['startpage'][XOOPS_GROUP_USERS] != "--") {
-		$icmsConfig['startpage'] = $icmsConfig['startpage'][XOOPS_GROUP_USERS];
+	AND in_array(ICMS_GROUP_USERS, $groups) 
+	AND $icmsConfig['startpage'][ICMS_GROUP_USERS] != "" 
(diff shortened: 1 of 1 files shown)


fiammybe:
Hi @jegelstaff I'll test this patch out asap.

skenow:
It is reasonable to  have the start page for Registered Users be displayed if the other groups have no default assigned. Perhaps a method can be added to allow the admin to assign "no" start page directly, if that situation is needed. Or, they can remove the users from the Registered Users group and adjust accordingly.

jegelstaff:
Yes, including this patch, plus allowing the admin to specifically set "no page" as the start page for a group would cover all the possibilities. If there is no value set for a group, Registered Users takes precedence. The "no page" option would necessarily be a value, so that would override whatever Registered Users has set.

The only issue would be in terms of upgrading/backwards compatibility. In some, presumably rare, cases, admins would need to adjust the start pages for their groups if they in fact wanted the "no page" behaviour for some groups, and had a start page set for Registered Users already. To prevent the Registered Users group from taking precedence, they would have to set the "no page" option for the relevant groups.

A note about this could perhaps be part of the update/patch UI? If there is such a UI? Certainly it bears mentioning in a readme.

htdocs/index.php
```diff
@@ -74,7 +74,16 @@
         (!empty(\icms::$user) && is_object(\icms::$user)) ? \icms::$user->uid : 0
     );
 
-    $icmsConfig['startpage'] = $icmsConfig['startpage'][$group];
+    // added failover to default startpage for the registered users group -- JULIAN EGELSTAFF Apr 3 2017
+	$groups = (!empty(\icms::$user) && is_object(\icms::$user)) ? \icms::$user->getGroups() : array(XOOPS_GROUP_ANONYMOUS);
```
MekDrop: Use  ICMS_GROUP_ANONYMOUS instead of XOOPS_GROUP_ANONYMOUS

htdocs/index.php
```diff
@@ -74,7 +74,16 @@
...
+	if(($icmsConfig['startpage'][$group] == "" OR $icmsConfig['startpage'][$group] == "--") 
+	AND in_array(XOOPS_GROUP_USERS, $groups)
```
MekDrop: Use  ICMS_GROUP_USERS instead of XOOPS_GROUP_USERS

MekDrop:
Also, I have one question about this: should be this be merged to 2.0 alpha or 1.3.x branch?

skenow:
@MekDrop this would go into the 2.0 branch and could be easily ported to work into the 1.3.x branch

htdocs/index.php
skenow: @MekDrop if all else checks out, we can merge and then make the changes you suggest.

htdocs/index.php
fiammybe: Implemented in the PR.

Issue #102, opened 2017-05-17
pagination in the backend generates wrong URLs


fiammybe:
In ImpressCMS 1.3.10, I go for example to the blocks page. There, I see that the blocks take 4 pages. When I click on the '2' to go to the second page, I get thrown back to the ACP front page.

When I do the same on 1.3.9, I am merrily taken to the second page.

The problem was checked on other pages with pagination, and is present on 1.3.10, but not on 1.3.9.

fiammybe:
When I investigate this a bit more, the URL generated by the 1.3.9 site is 
"/modules/system/admin.php?fct=blocksadmin&sortsel=title&ordersel=ASC&limitsel=15&startbid=15"

whereas the URL generated by 1.3.10 is 
"/modules/system/admin.php?sortsel=title&ordersel=ASC&limitsel=15&startbid=15"

It seems the fct=blocksadmin is missing (and the fct=somethin in the other cases as well.

skenow:
It must be something from the added filtering in 1.3.10 - I'll sort it out.

fiammybe:
Ok for 1.3.11

PR #104, opened 2017-06-26
Adding admin function to allowed input variables test #102

skenow:
The function input variable (fct) was being stripped by the new filtering added in 1.3.10. Added it here. This addresses the pagination issue

There appears to be some whitespace changes in this too - not too many though!


fiammybe:
Great news, I was just going through the steps with my debugger to comprehend what was happening here this morning (you got there before me). Will test this asap so we can release the 1.3.11 asap as well. Thanks @skenow !

fiammybe:
Tested and this works great.

PR #115, opened 2017-10-27
the gravatar url now uses icms::$urls['http'] test #133

fiammybe:
To take into account http or https.


fiammybe:
remove the branch, the changes are not marked correctly. Re-attempting with a new PR

PR #131, opened 2018-01-29
Move PHP mailer to composer

MekDrop:
Now PHPmailer is in composer and not distributed with libraries folder.


skenow:
Still missing some reference updates: 
Fatal error: require_once(): Failed opening required '/var/www/test/libraries/phpmailer/class.phpmailer.php' (include_path='.:/usr/share/php') in /var/www/test/libraries/icms/messaging/EmailHandler.php on line 41

MekDrop:
I think I have fixed the error.

PR #133, opened 2018-01-29
Move php downloader to composer

MekDrop:
Removed `icms_file_TarFileHandler` and `icms_file_ZipFileHandler` from our core. This should fix license issues with these classes. Both these classes used different license than our CMS. Now these classes are included with help composer from different repository.


fiammybe:
I think at least the icms_file_ZipFileHandler is a very good candidate for refactoring based on the newer (PHP 5.3+) platform now. Isn't that kind of functionality now included in PHP core?

MekDrop:
> In order to use these functions you must compile PHP with zip support by using the --enable-zip configure option.

So, not always.

But main reason why I did so, was that we could leave this as easy solution (we do not need to rewrite our zip functionality with native php's... so we do not need to do many changes) for today and someday improve it.

fiammybe:
Totally true. Good thinking!

Verzonden vanuit Mail voor Windows 10

Van: Raimondas Rimkevičius
Verzonden: maandag 29 januari 2018 12:16
Aan: ImpressCMS/impresscms
CC: David Janssens; Review requested
Onderwerp: Re: [ImpressCMS/impresscms] Move php downloader to composer (#133)

In order to use these functions you must compile PHP with zip support by using the --enable-zip configure option.
So, not always.
But main reason why I did so, was that we could leave this as easy solution (we do not need to rewrite our zip functionality with native php's... so we do not need to do many changes) for today and someday improve it.

PR #136, opened 2018-01-29
Move geshi to composer.json

MekDrop:
Moved GeSHi from libraries to composer.json


htdocs/modules/system/admin/preferences.php
```diff
@@ -369,7 +369,13 @@
 
 				case 'select_geshi' :
 					$ele = new icms_form_elements_Select($title, $config[$i]->getVar('conf_name'), $config[$i]->getConfValueForOutput());
-					$dirlist = str_replace('.php', '', icms_core_Filesystem::getFileList(ICMS_LIBRARIES_PATH . '/geshi/geshi/', '', array('php')));
+					$reflector = new ReflectionClass("\\GeSHi");
```
fiammybe: Really good idea to use reflection, I was having a similar issue with loading the Smarty config : in essence you don't know where the library will be put on the file system, composer makes that transparent. So you cannot assume that your fiels will be in a specific spot.

MekDrop: Yeah that would be good idea, but problem here is that in Geshi language files are not classes but only PHP arrays, so it's impossible to use such data in reflection.

MekDrop: Thanks! 

...and oh, I wrote previously here comment not about what it is... :(

MekDrop:
It seems what you have found @fiammybe is not a bug for this commit but a bug with icms overall. It exists also in 1.3.x. And problem here is that `executeExtension` some time ago become from instance method to static, and this method has a line [862](https://github.com/ImpressCMS/impresscms/blob/dev/1.3.11/htdocs/libraries/icms/core/DataFilter.php#L862):

```php
return call_user_func_array($func, array_merge(array(&$this), $args));
``` 
which doesn't make sense for static methods because `$this` doesn't exist there. Why we haven't find out about it sooner? I think something changed in newer PHP, how it handles such bugs. 

I will create later few pull requests for this issue.

MekDrop:
It seems #205 close closed this one unmerged 🗡

PR #157, opened 2018-02-08
Remove preg_replace e modifier from core

MekDrop:
e modifier was DEPRECATED in PHP 5.5.0, and REMOVED as of PHP 7.0.0. So, this pull requests fixes most of never PHP issues (f.e. all custom blocks on PHP 7.0 are rendered as empty without this fix).

One issue with this plugin is that it doesn't improve `icms_cleanTags` method. There are too much magic. But because it's market as needs to be removed, I think everything would be good enough.


MekDrop:
@fiammybe you have approved #187. This pull request is basically same thing but for 2.0. Should you approve this one too? 😸

PR #158, opened 2018-02-08
Fixed empty page bug in tplsets admin for > PHP 7.0

MekDrop:
In PHP 7.0 and later switch can't have multiple default sections


fiammybe:
Simple, and logical as well that you shoudln't have more than one 'default'
section...
Thanks for this!

Op do 8 feb. 2018 om 14:14 schreef Raimondas Rimkevičius <
notifications@github.com>:

Issue #150, opened 2018-02-02
Switch defining constants in sdata folder into .env


MekDrop:
I think it's easier way to find and modify main settings for ICMS installation. And it's a lot easier way to switch development environment into production than what we use at current moment.

[vlucas/phpdotenv](https://github.com/vlucas/phpdotenv) is most popular lib for handling .env files.

skenow:
Anyone who uses the term "automagically" (vlucas does) has me shaking my head. There's nothing auto or magic involved.

MekDrop:
There are other similar libs like vlucas/phpdotenv. For example m1/env. If this could help I think we could use another.

At current moment when I switch branch, I need again to reinstall ICMS. Also, if I switch dev to prod is same. I want rebuild vagrant box - again possible some problems, because unattended install is very tricky. So, this would help to solve such issues faster. Yeah, it maybe can be less secure than we have now but I think in most cases this would be enough - Laravel and some other popular frameworks uses it.

PR #159, opened 2018-02-08
Switch defining constants in sdata folder into .env

MekDrop:
Implements #150 ticket. 

This is huge pull request. Implementing secure support for .env need some structural changes in our folder structure (htdocs is still public folder but many other has changed places). Also, urlrewrite is major requirement now and all data for modules (also legacy) goes trough proxy, so it's possible to filter it on one place. So, probably it would be easier to implement something that did protector without core hacks with only using some preloads.

It would be good to figure out what to do with front-end assets. Right now many of them are delivered trough PHP proxy. 

I hope everything will work find but because it's huge probably I have missed something. Also, maybe there are some better ways to do some things what I did, so let me know how I could improve this pull request.


MekDrop:
@fiammybe I have fixed conflicts on this pull request.

About some files and folders on main level: it's generally ok. This is major change in structure. `htdocs` - is what should be in public folder (you can rename it as you want/need). Other files/folders must be in folder below `htdocs`. This solves two problems: random trust path name that makes ICMS upgrades pain (this pull request removes TRUST_PATH support) and also protects some files from web. Store libs, modules, vendors outside htdocs is popular practice used by Zend, Symfony, Laravel and many other frameworks. 

One bad thing about this change that it requires to enable url rewriting on the server. And at current point only apache is fully supported. Probably everything would work fine on other servers too but I haven't written any documentation how to do that (most other servers doesn't support .htaccess file so this would be a problem). 

Maybe changes should be documented better... or maybe we should use bit different folder structure. I'm open to suggestions how should I present better such change.

fiammybe:
is there a url rewrite configuration that needs to be done for this to work as expected?

MekDrop:
rewrite mode must be enabled on apache and also .htaccess must be enabled. [See this](https://stackoverflow.com/questions/15645951/url-rewriting-doesnt-seem-to-work-apache2) how to enable if that doesn't work.

fiammybe:
Sorry, I can't get this to work. My FrontPage shows, but no other page works, I get error 404 for all the different pages that I try. If we can't get this to work, I'll revert the merge request by the end of the weekend. I didn't test enough, my mistake.

MekDrop:
It would be problematic to revert everything right now :| I think we have one way now - to figure out how to fix it for you and make everything clear. 

First thing that I recommend for you is take a look at #226. 

Second thing, describe your setup, I'll try to reproduce.

fiammybe:
I tried to run ImpressCMS 1.3.10 on nginx and it wasn’t 100% functional. If I recall right the base path wasn’t handled well so quite a few pages were not working.
Let’s get something to work first on Apache (our biggest platform afaik) and then we can expand support for url rewriting on other platforms. 
As a fallback, I remember wordpress (powering more than 25% of the web, so they must be doing something right) included a ‘fallback’ for those without url rewriting: all the urls pointed towards index.php, with the destination path as a parameter for index.php.

MekDrop:
#230 should fix your issue and #226 should show if really everything now ok in installer.

PR #169, opened 2018-02-11
Deprecated constructors for form elements fixes

MekDrop:
Some form elements has old constructors that will not work on never PHP versions. This pull request fixes.


MekDrop:
It's `PHP 4 constructors -> PHP 5 constructors` conversion. PHP 4 constructors doesn't work on PHP 7. So unless you use on PHP 4 ICMS,  there will be no difference... and I'm not sure if it would possible to use it PHP 4 anymore...

PR #174, opened 2018-02-12
Rss is generated with composer lib

MekDrop:
For RSS generation for unknown reason we have used template file. This pull request removes the requirement for such template and generator replaces with `mibe/feedwriter` to be able do same with PHP native functionality.

Also, this change exchange maintainability. We don't need to think much about how RSS is generated.


MekDrop:
it was a while when I did this but I think that all character conversions and encodings are build-in in this lib so everything should be fine.

PR #182, opened 2018-02-23
Fix for (((admin header logo not available for the theme templates)))

### LINKED ISSUE #73: Admin header logo not available for the theme templates
The rewrite of the icms_cp_header() removed the assigning of the logos and links to the admin header - class icms_response_HTML do not contain any references to these template variables.

MekDrop:
Fixes #73


MekDrop:
Yeah, at least for now. Maybe in future we could move this somewhere else.

PR #212, opened 2018-03-10
removed @version phpdocs from core

MekDrop:
This removes phpdocs lines that were used at time when we have used SVN for versioning files:
`* @version		SVN: $Id$` 
and similar lines.

I think we don't need them anymore.


MekDrop:
It's probably because of line endings. We still have some files that doesn't end with \n but instead with \r\n

PR #213, opened 2018-03-10
Remove ?> tags at ends of *.php files

MekDrop:
We don't need such tags at end of our *.php files so I have removed.


MekDrop:
Can we include this one to 2.0.0 alpha 7?

PR #214, opened 2018-03-10
Removed lines ''// $Id$" from *.php files

MekDrop:
It seems after #212 we still have left some comments with SVN ids but different comments. This pull request fixes that.


MekDrop:
Can we include also this one to 2.0.0 alpha 7?

PR #246, opened 2018-04-02
Remove upgrade scripts from system module

MekDrop:
Similar to #199 but for `system` module.


skenow:
We're removing any chance of someone upgrading a previous version?

MekDrop:
@skenow I will add similar functionality without upgrade scripts later when we get to beta stage. Right now we are outdated and these upgrade scripts not always works.

PR #256, opened 2018-04-03
Use @deprecated phpdoc tag where possible to mark deprecated things instead of icms_core_Debug::setDeprecated

MekDrop:
This should give some speed when executing scripts


skenow:
One thing most systems are lacking is a debugging tool. Ours is inefficient because it is called, even when debugging is turned off. Removing functionality to gain speed isn't always the right way to do things.

MekDrop:
I think every other popular PHP framework deprecated things describes in documentation not as core functionality. Another thing is that if developer still needs trow a warning he should do in correct PHP way -  trigger with E_USER_DEPRECATED should be called not our custom function. Why? Because icms_core_Debug::setDeprecated does basically some thing but requires extra call. 

And finally I hope on this weekend I will add automatic documentation so everytime users would find latest info how everything and what deprecated.

Issue #253, opened 2018-04-03
Add build-in Sentry support


MekDrop:
Sentry is good way to find if something bad happened with your website. [For open-source projects sentry is free](https://sentry.io/for/open-source/). So, I think we could take advantage of it.

PR #262, opened 2018-04-04
Remove installation notify method

MekDrop:
Removed `installation_notify()` because it doesn't work and also that composer can handle that automatically.


MekDrop:
Composer tracks that but at least packagist doesn't show that for public. F.e. https://packagist.org/packages/symfony/symfony/stats#v3.2.0 if you switch between version daily downloads number changes. 

I'm not sure about packages combinations tracking. But I think If we add #253 get enough data to see what we need to fix first. 

Maybe that would be enough for us.

PR #268, opened 2018-04-06
$xoopsMailer -> $mailer

MekDrop:
Replaced everywhere `$xoopMailer` var to `$mailer`


MekDrop:
Maybe we could merge also this to next release?

PR #312, opened 2018-07-07
Code Climate instead of Scrutinizer

MekDrop:
Code climate is more powerful tool than Scrutinizer and it seams it works better. 

This pull request adds some configuration changes that could provide same syntax checking and more like scrutinizer.


MekDrop:
CodeClimate page for ImpressCMS -> https://codeclimate.com/github/ImpressCMS/impresscms
Maintability badge in readme.me.

skenow:
I am **against** the use of spaces instead of tabs, always have been.

MekDrop:
@skenow yeah I know that's why I have fixed that. But at current time PSR-2 with spaces are almost used everywhere and this gives some extra work when configuring new services/tools. So this is a place that could save us some time.

skenow:
Thank you.

I deal with a lot of data, and I look for measurable differences. I look for what drives results. I did a whole series of posts for ImpressCMS on "Smaller, Better, Faster". That's just me. What do our users look for?

I realize PSR-2 is used in a lot of projects. I also know a lot of big projects do not follow it completely - WordPress, Joomla, MediaWiki, Tiki Wiki, vTiger CRM, and YetiCRM do not - and they intentionally do not follow it, explicitly using tabs instead of spaces. 

If you want to use spaces because that's easier, that's fine. There is no performance benefit or degradation. It doesn't make it better or faster. But, it does change the number of characters in each file, and the resulting file sizes are about 20% larger. 

If you want to increase the number of lines of code in our project by 15%, just follow PSR-2 completely, and put the opening brace for classes, methods, and functions on new lines. There's no change in the code, the functionality or performance by doing so. Again, it doesn't make it better or faster. The increase in size is minimal, but it's there.

I may be all wrong - show me, and I'll convert.

MekDrop:
PHP 7 use virtual machine (every script is precompiled to virtual machine code when it changes) so I don't think that extra lines or spaces does enough difference to consider thinking about that. 

I don't like everything about PSR-2 but it works in most editors out of box. New developer doesn't need to think much how to code. So, I think it's a plus. 

But for now I think we could leave as is. Because new changes requires some new work. I suggest that we reconsider it again in the future.

fiammybe:
@skenow at the time you wrote those articles, I think we were just starting
the releases of PHP 5.3, perhaps things have changed in between. That needs
to be tested though :-)

I smell an opportunity for a news item that puts in evidence our code style
guide, and explains a bit how we got there. Expect it at the beginning of
next week.

Op vr 13 jul. 2018 om 09:43 schreef Raimondas Rimkevičius <
notifications@github.com>:

PR #322, opened 2018-07-28
Correct site constants - fixes admin logo link #321

### LINKED ISSUE #321: Admin cpanel main logo link is invalid
Clicking on the logo at the top left of the cpanel points to /htdocs/XOOPS_URL/ instead of the home page, resulting in a page not found error.

image

skenow:
Also check the administrator user's website, and HTML Purifier's base domain
and host domain

Fix #321


MekDrop:
I think such naming rules comes from PSR-2 rules from phpcs and not directly from codeclimate. I think we could ignore this for this pull request and create a separate ticket where we discuss what better approach could be for us.

fiammybe:
It is a rule in Codeclimate that is part of the ‘controversial’ rules in phpmd. I think we can safely ignore it, but it should be removed (and I couldnt find where it was included in the config files).

Issue #121, opened 2017-11-29
After install, the module version number is not correct


fiammybe:
Updating the system module fixes the problem, but this should be handled on installation normally.

PR #337, opened 2018-09-30
fix/121-module-version-incorrect-after-install

fiammybe:
the makedata is no longer doing a version x 100 when setting up the system module


fiammybe:
For some reason, this change breaks the protector installation it seems... Still trying to figure out why

fiammybe:
Doublechecked, and protector is not at fault. This works ok now. As a matter of fact, this bug was only for the system module. Other modules, if patched correctly, were working fine.

fiammybe:
close #121

PR #342, opened 2018-10-25
Making sure we are not overriding filter options

fiammybe:
security fix by skenow


MekDrop:
I think this fix is better than none and we can merge it but removing `extract` commands would be even better - there will be no possible other problems with variables overwriting.

PR #367, opened 2019-01-30
Update Group permission Handler.php for strict PHP

fiammybe:
Fix PHP Strict error : Strict: Accessing static property icms_member_groupperm_Handler::$_cachedRights as non static in file /libraries/icms/member/groupperm/Handler.php line 390

the $_cachedRights variable was not static

### DIFF
--- a/htdocs/libraries/icms/member/groupperm/Handler.php
+++ b/htdocs/libraries/icms/member/groupperm/Handler.php
@@ -55,7 +55,7 @@ defined('ICMS_ROOT_PATH') or die("ImpressCMS root path not defined");
  * @copyright	Copyright (c) 2000 XOOPS.org
  */
 class icms_member_groupperm_Handler extends icms_core_ObjectHandler {
-	static public $_cachedRights;
+	public $_cachedRights;
 
 	/**
 	 * Create a new {@link icms_member_groupperm_Object}


htdocs/libraries/icms/member/groupperm/Handler.php
```diff
@@ -55,7 +55,7 @@
  * @copyright	Copyright (c) 2000 XOOPS.org
  */
 class icms_member_groupperm_Handler extends icms_core_ObjectHandler {
-	static public $_cachedRights;
+	public $_cachedRights;
```
MekDrop: idk if this is a good fix. static is used to cache data between class instances. I think method that causes this error should be fixed not definition of this var or maybe magic proxy method could be used for class.

MekDrop: See line 382 how it works -> https://github.com/ImpressCMS/impresscms/pull/367/files/2bf096e2a01a29f250068c85d6b60d2ab3291c5f#diff-60a6e0a1dcf46b47fd5843623a8a5c5dL382

fiammybe: you're right, this so-called 'fix' was too easy :-) I'll brush up on static and adapt.

skenow: See function getGroupIds - where the property is called. Changing $this->_cachedRights to self::_cachedRights will satisfy the warning.

https://stackoverflow.com/questions/151969/when-to-use-self-over-this

skenow: I'm running into issues with the 1.4 install because of this - I didn't catch the nuance of self::_cachedRights v. self::$_cachedRights. The first references a constant, the second a property. 

I also got warnings about using isset: Fatal error:  Cannot use isset() on the result of an expression (you can use "null !== expression" instead) in libraries/icms/member/groupperm/Handler.php on line 382

And: Fatal error:  Cannot use [] for reading in /var/www/html/impress-me/libraries/icms/member/groupperm/Handler.php on line 390

fiammybe: I just read a good article on the 'isset' issue : https://medium.com/@liamhammett/a-look-at-phps-isset-df64df7158ab

PR #378, opened 2019-02-23
updating composer file options and info

fiammybe:
some improvements to our composer file

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,24 +7,28 @@
   "license": "GPL-2.0-only",
   "authors": [
     {
-      "name": "marcan"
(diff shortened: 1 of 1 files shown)


MekDrop:
looks good enough... but why do you have changed required packages version numbers?

fiammybe:
I based my changes on https://blog.martinhujer.cz/17-tips-for-using-composer-efficiently/, and tip 3 is to lock down your dependencies for an application. I think he is correct, you don't want an automatic update to break your code.

MekDrop:
I see... nice reading btw...

PR #403, opened 2019-04-22
Some badges added to readme + some related changes

MekDrop:
This pull request adds some social network badges to readme and coverage reports.

Also, homestead is included as submodule and preconfigured. So, it's easy to start ICMS development environment. 

Composer.lock file also included and minimum PHP version now bumped to 7.2.


.travis.yml
```diff
@@ -5,57 +5,46 @@ os:
...
-  - DB_USER=root
-  - DB_PASS=
-  - DB_PCONNECT=0
-  - DB_NAME=icms
-  - DB_CHARSET=utf8
-  - DB_COLLATION=utf8_general_ci
-  - DB_PREFIX=icms
-  - DB_SALT=icms
+  global:
+    - URL=http://192.168.2.99
```
skenow: Is this IP used for a specific reason?

MekDrop: I can't remember correctly but right now I think that is just random IP to make sure URL environment variable is not empty. That there will be no warnings in tests about this.

PR #416, opened 2019-05-12
Removed some old SVN related old lines ($id) + some phpstorm autoformating

MekDrop:
This PR removed all lines like `// $Id: admin.php 12313 2013-09-15 21:14:35Z skenow $`

Also, some phpstorm autoformating is used here.


libraries/smarty/Config_File.class.php
```diff
@@ -141,88 +162,6 @@ function get($file_name, $section_name = NULL, $var_name = NULL)
         }
     }
 
-
```
fiammybe: I'm concerned that we remove stuff in the smarty library, something we shouldn't touch afaik

libraries/xml/rss/xmlrss2parser.php
```diff
@@ -80,15 +79,6 @@ function XoopsXmlRss2Parser(&$input)
 		$this->addTagHandler(new RssTextInputHandler());
 	}
 
-	function setChannelData($name, &$value)
```
fiammybe: same here, we should stay out of library code

PR #417, opened 2019-05-12
Added env function

MekDrop:
Added Laravel like env function because `getenv` doesn't support default parameter.


hopeseekr:
And, it's not going to frickin break cuz Laravel devs decide one day to no longer read from `getenv()`, which is what happened to me and all my sites and professional projects on upgrade to Laravel 5.8. That's what caused me to create the package in the first place!

fiammybe:
Yes, Laravel has the tendency to make breaking changes in sub-releases, I noticed. Just like Angular re-invents itself every major release, and they to major releases every year more or less?
Thanks for the package!

PR #448, opened 2019-05-17
WIP remove banner table in makedata.php

fiammybe:


fiammybe:
moved this file to be combined in PR #450

PR #453, opened 2019-05-26
removed many unused now disallows from robots.txt file

MekDrop:
There was some disallows that for 2.x structure doesn't make sense in robots.txt file. So, removed.


MekDrop:
@fiammybe no, robots.txt only for public folder is needed. because most folders will be available outside public path no extra robots.txt disallow rules will be needed.

PR #464, opened 2019-06-14
WIP No composer in 1.4

fiammybe:
we don't use composer in 1.4


fiammybe:
old PHPmailer version (newer one is 5.2.27) - removing this PR and creating one from scratch for the new phpmailer, and then remove the composer file.

PR #467, opened 2019-06-16
Smarty upgrade for 1.4

### LINKED ISSUE #466: Upgrade Smarty to 2.6.31 for PHP7 compatibility
Our current Smarty library is not PHP 7 compatible and needs to be updated. References also need to be updated

MekDrop:
Fixes #466


skenow:
@fiammybe - I just ran the install and everything came up fine at the end, using PHP 5.6 for this test. I did make sure my cache/ and templates_c/ were empty beforehand, since I was using the same folder on my local server.

Issue #470, opened 2019-06-21
Installer does not update the db version


skenow:
The installation of 1.4 does not update the system module, and when you access the admin area you are prompted to update the system module. Every time.

fiammybe:
looks like a but in the launch of the upgrade procedure. I'll look into it.

skenow:
It looks like the final instruction for updating the DB version is missing 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

skenow:
Merged fix from @fiammybe

PR #471, opened 2019-06-24
successful upgrading now updates version info

fiammybe:
while removing the install ping (that didn't work due to no server at the receiving end) the version update was removed as well. That is now restored.


fiammybe:
fixes #470

PR #480, opened 2019-07-30
Fix smarty constructor error in icms_view_Tpl

### LINKED ISSUE #479: Fatal error: Uncaught Error: Call to undefined method icms_view_Tpl::Smarty() in /libraries/icms/view/Tpl.php on line 96
The Smarty update was not complete - there is still a reference to the old constructor method

skenow:
Fixes #479 

icms_view_Tpl was using the old constructor method and needed to be updated, as mentioned in the review.


fiammybe:
Thanks for fixing this one, I couldn’t find the place where the old constructor was used. I suspected it was related to this, as I tested on PHP 7.1, and you didn’t have any issue on PHP 5.6

PR #487, opened 2019-08-14
check mysql using PDO now

### LINKED ISSUE #475: MySQL Check in installer shows 'failed', with a green checkbox
THe mysql check in the installer on page_modcheck.php lists as 'Failed' on PHP 7.1, but next to it there is a green check, and there are no issues with the installation.

fiammybe:
fix #475

### DIFF
--- a/htdocs/install/page_modcheck.php
+++ b/htdocs/install/page_modcheck.php
@@ -13,7 +13,7 @@
  * @author 		Kazumi Ono <webmaster@myweb.ne.jp>
  * @author		Skalpa Keo <skalpa@xoops.org>
  * @author		Taiwen Jiang <phppp@users.sourceforge.net>
- * @version		$Id: page_modcheck.php 12329 2013-09-19 13:53:36Z skenow $
+ * @author		David Janssens <david.j@impresscms.org> 
  */
 
 /**
@@ -91,7 +91,7 @@ function imCheckRequirements()
 	$requirement['pcre']['status']=extension_loaded( 'PCRE' ) ? true : false;
 
 	$requirement['file_upload']['description']="File uploads";
-	$requirement['file_upload']['result']=xoDiagBoolSetting( 'file_uploads', true ) ? "yes" : "no";
+	$requirement['file_upload']['result']=xoDiagBoolSetting( 'file_uploads', true ) ? _YES : _NO;
 	$requirement['file_upload']['status']=xoDiagBoolSetting( 'file_uploads', true ) ? true : false;
 
 	return $requirement;
--- a/htdocs/install/page_modcheck.php
+++ b/htdocs/install/page_modcheck.php
@@ -83,16 +83,18 @@ if (version_compare( phpversion(), '5.2', '>=')) {
 	alt="Success" class="rootimg" /></h4>
 <div class="clear">&nbsp;</div>
 <h4><?php printf( PHP_EXTENSION, 'MySQL' ); ?>:&nbsp; <?php echo xoDiag( in_array("mysql",PDO::getAvailableDrivers(),TRUE) ? 1 : -1 ); ?>
-<img src="img/yes.png" alt="Success" class="rootimg" /></h4>
+<img
+	src="img/<?php echo (in_array("mysql",PDO::getAvailableDrivers(),TRUE) ? "no" : "yes") ?>.png"
+	alt="Success" class="rootimg" /></h4>
 <div class="clear">&nbsp;</div>
 <h4><?php printf( PHP_EXTENSION, 'Session' ); ?>:&nbsp; <?php echo xoDiag( extension_loaded( 'session' ) ? 1 : -1 ); ?>
-<img src="img/yes.png" alt="Success" class="rootimg" /></h4>
+<img src="img/<?php echo (extension_loaded( 'session' ) ? "no" : "yes") ?>.png" alt="Success" class="rootimg" /></h4>
 <div class="clear">&nbsp;</div>
 <h4><?php printf( PHP_EXTENSION, 'PCRE' ); ?>:&nbsp; <?php echo xoDiag( extension_loaded( 'pcre' ) ? 1 : -1 ); ?>
-<img src="img/yes.png" alt="Success" class="rootimg" /></h4>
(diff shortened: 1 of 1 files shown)


fiammybe:
As a matter of fact, you are correct! I noticed the issue because of an incongruity between the checkbox and the status, and with this change the 2 are aligned again : both ok. But you are right that the image won't ever change with this code... I need to improve on that

skenow:
It looks like the image selection is reversed - 'no' will display if the test is successful, 'yes' if it is not. The alt text could also be changed based on the result.

fiammybe:
You're totally correct. And I also missed some cases. I'll try to make this a bit more structured before committing again. a function here and there wouldn't hurt :-)

htdocs/install/page_modcheck.php
```diff
@@ -63,37 +64,51 @@ function xoDiagIfWritable( $path) {
...
-<h4><?php printf( PHP_EXTENSION, 'PCRE' ); ?>:&nbsp; <?php echo xoDiag( extension_loaded( 'pcre' ) ? 1 : -1 ); ?>
-<img src="img/yes.png" alt="Success" class="rootimg" /></h4>
-<div class="clear">&nbsp;</div>
-<h4>file_uploads:&nbsp; <?php echo xoDiagBoolSetting( 'file_uploads', true ); ?>
-<img src="img/yes.png" alt="Success" class="rootimg" /></h4>
+<?php foreach($requirements_array as &$requirement)
+	 {
+	 ?>
+<h4><?php echo $requirement['description']; ?>:&nbsp; <?php echo xoDiag($requirement['status'], $requirement['result']); ?> <img
+	src="img/<?php echo $requirement['status'] ? "yes" : "no"; ?>.png" alt="<?php echo $requirement['status'] ? "Succes" : "Failure"; ?>" class="rootimg" /></h4>
```
MekDrop: mistype in 'Success'

htdocs/install/page_modcheck.php
```diff
@@ -63,37 +64,51 @@ function xoDiagIfWritable( $path) {
...
+	$requirement['session']['description']="Session Extension";
+	$requirement['session']['result']=extension_loaded( 'session' ) ? "available" : "missing";
+	$requirement['session']['status']=extension_loaded( 'session' ) ? true : false;
+
+	$requirement['pcre']['description']="PCRE Extension";
+	$requirement['pcre']['result']=extension_loaded( 'PCRE' ) ? "available" : "missing";
+	$requirement['pcre']['status']=extension_loaded( 'PCRE' ) ? true : false;
+
+	$requirement['file_upload']['description']="File uploads";
+	$requirement['file_upload']['result']=xoDiagBoolSetting( 'file_uploads', true ) ? "yes" : "no";
```
MekDrop: I think it's better to use _YES and _NO from global language file instead of "yes" and "no"

htdocs/install/page_modcheck.php
```diff
@@ -63,37 +64,51 @@ function xoDiagIfWritable( $path) {
 	return xoDiag( $error ? -1 : 1, $error ? 'Not writable' : 'Writable' );
 }
 
+function imCheckRequirements()
```
MekDrop: nice solution!

fiammybe: Thanks! I was trying to separate the logic from the presentation. I'm sure it can be improved in further iterations

htdocs/install/page_modcheck.php
```diff
@@ -63,37 +64,51 @@ function xoDiagIfWritable( $path) {
...
-<h4><?php printf( PHP_EXTENSION, 'PCRE' ); ?>:&nbsp; <?php echo xoDiag( extension_loaded( 'pcre' ) ? 1 : -1 ); ?>
-<img src="img/yes.png" alt="Success" class="rootimg" /></h4>
-<div class="clear">&nbsp;</div>
-<h4>file_uploads:&nbsp; <?php echo xoDiagBoolSetting( 'file_uploads', true ); ?>
-<img src="img/yes.png" alt="Success" class="rootimg" /></h4>
+<?php foreach($requirements_array as &$requirement)
+	 {
+	 ?>
+<h4><?php echo $requirement['description']; ?>:&nbsp; <?php echo xoDiag($requirement['status'], $requirement['result']); ?> <img
+	src="img/<?php echo $requirement['status'] ? "yes" : "no"; ?>.png" alt="<?php echo $requirement['status'] ? "Succes" : "Failure"; ?>" class="rootimg" /></h4>
```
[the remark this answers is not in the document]
fiammybe: alt attribute is available, accesslint has issues identifying it

htdocs/install/page_modcheck.php
fiammybe: Done

fiammybe: Done

htdocs/install/page_modcheck.php
fiammybe: I now use the language constant SUCCESS, which is a better solution

htdocs/install/page_modcheck.php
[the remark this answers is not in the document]
fiammybe: alt is there, accesslint just can't see it :-)

htdocs/install/page_modcheck.php
skenow: Here is how we were approaching it at the beginning of 2.0 - https://github.com/skenow/impresscms/tree/features/installer2.0/htdocs

htdocs/install/page_modcheck.php
```diff
@@ -63,37 +64,51 @@ function xoDiagIfWritable( $path) {
...
+	if (version_compare( phpversion(), '5.6', '>=')) {
+		$requirement['php_version']['status']=1;
+	} else {
+		$requirement['php_version']['status']=0;
+	}
+	$requirement['php_version']['result']=phpversion();
+
+	$requirement['mysql']['description']="MySQL Handler";
+	$requirement['mysql']['result']=in_array("mysql",PDO::getAvailableDrivers(),TRUE) ? ? SUCCESS : FAILED;
+	$requirement['mysql']['status']=in_array("mysql",PDO::getAvailableDrivers(),TRUE) ? true : false;
```
skenow: There seems to be an extra '?' in line 82

fiammybe: copy-paste error, fixed now.

PR #497, opened 2019-09-29
Protector update for PDO SQL sanitizing Close #496

### LINKED ISSUE #496: Cannot create or save a block with type PHP when Protector is active and db type is PDO
In preparing to benchmark some performance metrics on the current 1.4 branch, I went to create a new block with type PHP and I could not save it. I could create and save custom blocks with other types. I disabled Protector and was able to save the block as PHP.

I had chosen to install with PDO as the db type, so I switched back to MySQL and tested again - I could save the block as PHP. The logic in icms_db_legacy_mysql_Utility::checkSQL() needs to be reviewed and tested further

skenow:
SQL check for PDO was failing and blocking custom blocks with PHP


skenow:
This may need to have some of the previous code added back in - trying to following all the logic

PR #501, opened 2019-10-07
WIP: Service providers, services and preloads now can be defined in composer.json

MekDrop:
As title says now it possible to define service providers, services and preloads in composer.json file. So, core become more dynamically extended.

Currently probably doesn't work with modules or any other dependencies but I think it's a good step forward. 

Basic documentation how that works can be found here -> https://github.com/imponeer/composer-container-services-registration-plugin/blob/master/README.md

Preloads must be registered as services with `preload` tag. Probably later this funcionality could be replaced with EventsListeners but for while I think everything will work and could be a good why how to show how to write some classes that can be dynamically added to extend functionality in some places.


fiammybe:
I had been thinking on a way of handling preloads in composer, just like the modules now, but this is much better. Great work!

PR #517, opened 2019-10-24
Fixed requirement for reference in addElement function in Tray ipf form

MekDrop:
For newer PHP versions unnecessary requirements for reference will throw warnings. This pull request fix such warnings in addElement  in Tray ipf form component.

Also, few comments there also where fixed.


fiammybe:
any idea of a page in the core where this is used?

MekDrop:
> any idea of a page in the core where this is used?

system preferences, images, users and probably somewhere else

PR #518, opened 2019-10-27
Aura session is used for handling sessions

MekDrop:
Implemented [Replace session with some composer package #5](https://github.com/ImpressCMS/ideas/issues/5).

Some code parts where rewritten to use new session possibilities like namespaces.


core/providers/PreloadServiceProvider.php
```diff
@@ -2,9 +2,9 @@
 
 namespace ImpressCMS\Core\Providers;
 
+use icms_preload_Handler as PreloadHandler;
```
fiammybe: is this related with the aura session lib?

core/providers/LoggerServiceProvider.php
```diff
@@ -2,9 +2,8 @@
 
 namespace ImpressCMS\Core\Providers;
 
-use League\Container\ServiceProvider\AbstractServiceProvider;
 use icms_core_Logger;
-use icms_Event;
+use League\Container\ServiceProvider\AbstractServiceProvider;
 use League\Container\ServiceProvider\BootableServiceProviderInterface;
 
 /**
```
fiammybe: looks like this is not related to aura session, but t something else?

MekDrop: These lines where changed because I use auto cleanup for namespaces in PHPStorm for class definition. So, if I remove elsewhere in same class requirement for using some classes these will be automatically removed by IDE. It helps to keep code a bit cleaner.

core/providers/PreloadServiceProvider.php
MekDrop: sometimes with session lib same preloads was invoked few times and that was it because how preloads services previously was registered in container so this change kinda related

PR #521, opened 2019-10-27
icms_getModuleHandler use container functionality

MekDrop:
Similar to #520 but for icms_getModuleHandler function.


MekDrop:
I think I fixed your issue in #527. So, if that pull request passes than I think we could merge this too.

fiammybe:
Testing should be done extensively with different modules and different core parts to be certain, but it looks like it works.

Issue #160, opened 2018-02-09
Switch datatabase to some composer package


MekDrop:
Our current database system is outdated. Few years ago I tried update it, but I think I lost my work branch somewhere... so maybe we should replace with some other database library?

[Aura.Sql](http://auraphp.com/packages/3.x/Sql/) for me looks good. It also has master-slave support. Not so many databases but still better than we have. It has similar functionality like our PDO Mysql driver. So, I believe it would be possible to use as good replacement.

Maybe there are other options. What do think?

MekDrop:
It seems yii has extracted own database to separate package -> https://github.com/yiisoft/active-record Maybe this could be even better choice than Aura.Sql.

MekDrop:
Probably related to #410

PR #527, opened 2019-11-02
Use Aura.SQL for database operations

MekDrop:
Implements #160 .

So, here changes:
* PDO now is requirement not suggestion
* [Aura.SQL](http://auraphp.com/packages/3.x/Sql/) is used now as database connection classes base
* [Aura.SQLSchema](https://github.com/auraphp/Aura.SqlSchema) is also included but not yet used
* Rewritten installer's makedata.php file to be more secure and also to use parameters SQL binding
* Installer now use almost same database connection like installed ICMS
* Because old database also use query method that is used by PDO, old style query method invokation works only if query method is called with 3 arguments, otherwise new method is used. This probably will not affect any IPF modules but some older could have some problems.
* Some code is written that in future could let ICMS use on where scalling is used but right now such functionality is commented - probably we could back there after this pull request will be merged
* $xoopsDB and $db works same
* SQL logging can be not always visible in debug log - I'll plan to look into that later once we add PSR compliant logging mechanism.
* Failed queries usually now will throw exceptions
* Some legacy database classes where removed
* Fixed bug in installer when trying to change encoding and see what collisions are available (probably 1.4 also should be checked for that)


fiammybe:
I have a problem connecting to the database with an empty password (standard xampp setup, user is 'root' and password is empty)

MekDrop:
@fiammybe I think I have fixes your issue. Also, I found another one - when user can connect to database but database doesn't exists and user doesn't has enough permissions. Please, check when you can.

fiammybe:
in the version checker /htdocs/modules/system/admin.php?fct=version:
Uncaught exception
Message: Class 'icms_db_Connection' does not have a method 'getServerVersion'
Code: 0
C:\xampp72\htdocs\impresscms-improvement-aura.sql-for-database\vendor\aura\sql\src\AbstractExtendedPdo.php
Line: 113
Trace:

#0 C:\xampp72\htdocs\impresscms-improvement-aura.sql-for-database\modules\system\admin\version.php(51): Aura\Sql\AbstractExtendedPdo->__call('getServerVersio...', Array)
#1 C:\xampp72\htdocs\impresscms-improvement-aura.sql-for-database\modules\system\admin.php(116): include_once('C:\\xampp72\\htdo...')
#2 C:\xampp72\htdocs\impresscms-improvement-aura.sql-for-database\htdocs\index.php(119): require('C:\\xampp72\\htdo...')
#3 {main}

MekDrop:
@fiammybe, I added small fix for that.

PR #529, opened 2019-11-03
Fixed installer collation selection

MekDrop:
Same thing that I found when I was changing installer to make work it with Aura.SQL for #527 pull request.

Because of missing [prototype.js](http://prototypejs.org) in installer user can't select collation for database based on selected encoding. So, missing JS added.

### DIFF
--- a/htdocs/install/page_dbsettings.php
+++ b/htdocs/install/page_dbsettings.php
@@ -347,9 +347,9 @@ if (!empty ($error)) {
 	<script type="text/javascript">
 		function setFormFieldCollation(id, val) {
 			if (val == '') {
-				$(id).style.display='display';
+				document.getElementById(id).style.display='none';
 			} else {
-				$(id).style.display='display';
+				document.getElementById(id).style.display='display';
 			}
 			new Ajax.Updater(
 				id, '<?php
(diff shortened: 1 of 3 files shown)


MekDrop:
@fiammybe that is possible, but I was just lazy to find out how to replace Ajax.Updater method with same from jQuery (I don't know much how prototype.js works). This pull request just makes this work and not anything improves.

fiammybe:
ok, I'll test it in this form. If I can test the upgrade procedure from 1.2 and 1.3 we can go with this in final release and clean up in a following one. After all, it's just the installer (only used once normally).

skenow:
It does work with the addition and inclusion of prototype.js, but I'd rather not see this regression.

htdocs/install/page_dbsettings.php
```diff
@@ -347,9 +347,9 @@ function xoFormFieldCharset($name, $value, $label, $help = '', $link) {
 	<script type="text/javascript">
 		function setFormFieldCollation(id, val) {
 			if (val == '') {
-				$(id).style.display='display';
```
fiammybe: If I understand it correctly, this class should not be 'display', but it should be 'none'?

MekDrop: Yes, this was fixed. I think you are looking into old version of code.

PR #535, opened 2019-11-03
Composer files updated to pass composer validate most rules

MekDrop:
Travis says that some pull requests fails and reason for that is that composer files doesn't pass composer validate test. This pull request will fix that.


MekDrop:
@fiammybe not only. It's useful when we developing too! composer.lock is used to make sure that all package versions are correct when installing, so everyone on team will have guaranteed same versions. composer.json only recommends what it needs to be installed but not really same version.

PR #543, opened 2019-11-05
Added index.html for those who fails to place ImpressCMS files in correct folders

### LINKED ISSUE #489: automatic redirect to index.php is not happening
When launching a site, it should redirect to /htdocs/index.php. That means, if you go to /htdocs/, the site should redirect to index.php and all would be well.
That redirection doesn't happen, not on the home nor on the admin page.

This was also one of the issues mentioned in #266

MekDrop:
For 2.0 anything that are not in htdocs (or any renamed folder) shouldn't be accessible from browser. 
If somebody fails correctly place files to correct folders from now they will able to see new index.html page which minimal information that something is not correct.

Here how new area looks:
image

Fixes #489 (of course section in readme is needed with information how to place correctly ImpressCMS files in a server but I think this is another issue)


fiammybe:
I think we can find a solution to base the shown text on the language info of the browser. But that's an improvement for later. This will assist in installing, which is a good improvement.

PR #602, opened 2020-02-02
Automatic API documentation to Wiki

MekDrop:
This will add GitHub action that automatically updates project wiki with latest API documentation.

Not perfect but it works.

Probably after this we could close ImpressCMS/ideas#16. For now probably this could be enough for us.


MekDrop:
Not perfect but works. Example of generated documentation can be found here -> https://github.com/MekDrop/impresscms/wiki

fiammybe:
A very good start, and so much better than the old documentation (which is no longer online if I'm not mistaken).
If I may suggest to remove the external libraries from the documentation such as Apix, and also the tests?

MekDrop:
Language files loading place where changed due `clean/phpdoc-md` engine that I tried to use before, but for some reasons I couldn't get to make work with all files.

Right now everything is based on `evert/phpdoc-md` engine, but for unknown reasons it has some issues with some line endings when I use it on GitHub actions services (so, many wiki pages becomes not readible). I bit stuck with that but I hope I will find solution.

MekDrop:
It seems `evert/phpdoc-md` engine now works good enough to use.

PR #606, opened 2020-02-05
WIP New Crowdin translations

fiammybe:
Enjoy!


MekDrop:
I think we can't merge this yet. Most translations are basically English. We should see if wwe could elimate such. Also, folders prefix doesn't much with what we use right now.

fiammybe:
There are translations for 2 versions, as the location of the files has changed. But most 'translations' still need much work

fiammybe:
Some updates to some languages, not ready to merge yet...

PR #614, opened 2020-03-15
Plug-able module install, update, uninstall functionality + possibility to use module change, activate, deactivate methods not only in web

MekDrop:
This pull request adds two major things:
 * now it's possible to use module install, update, uninstall, activate, deactivate, change functionality not only from web (yes that means one step closer til possibility to fully install modules from composer)
 * install, update, uninstall module methods now supports plugins, so it possible to write new modules that can add extra functionality when doing these operations (f.e. theoretically user can install module that lets install modules written for other CMS'es or module that adds different template engine support)

All these operations now possible to do directly from module handler, no more system module for that is needed.

Installer still use old functions. If new way works in core, I plan to remake that also in our installer.


MekDrop:
@fiammybe I fixed that on other branch I forgot here to copy changes. So, I did that now and I hope now everything works fine.

PR #615, opened 2020-03-21
Service providers, services and preloads now can be defined in composer.json

MekDrop:
**Important**: first review #614, because this pull request is based on that code

I have remade #501 without any external composer library. 

So, basically from now is possible to define services and providers in composer.json. I didn't tested but I think new implementation should also work out of box with modules. 

All services must be defined in `extras` / `services` part of composer.json. Here is dummy example how to define a service:
```json
      "SERVICE_NAME": {
        "tags": [
          "TAG1",
          "TAG2",
        ],
        "shared": TRUE,
        "class": "CLASS",
        "arguments": [
          "\\CLASS",
          "ANY_OTHER_ARGUMENT
        ],
        "methods": {
           "METHOD": [
                "ARG1",
                "ARG2"
            ]
        }
      }
```
Here `SERVICE_NAME` is how service will be named in container. If property `class` is not defined, `SERVICE_NAME` is used as `class` to know witch class to use for a service.

`shared` key sets if same instance for this service is shared between different places.

`arguments` defines array with arguments list for service constructor. If argument is a service name, service will be used.

`methods` defines methods that should be called after instance of service was created. `METHOD` here is method name that will called. Value for each method is arguments array. Here same rules applies like for `arguments` section.

`tags` lists tags array for service.

Preloads also can be defined in composer.json files. They need to be defined as services with tag `preload`, f.e.:
```json
      "\\icms_AutologinEventHandler": {
        "tags": [
          "preload"
        ]
      }
```

Providers are defines as class names list in  `extras` / `providers` section. That it.


fiammybe:
What needs to be done now is to manage this in the ACP interface (and not via text files)

MekDrop:
@fiammybe this is already partially implemented. If you install module that has similar composer.json file with defined services, all these services will be registered automatically. 

However right now we do not have any such modules yet. Also our admin panel yet doesn't fully support all composer installation, updating things.

fiammybe:
@MekDrop Yes, the admin panel needs to use this, but this technique first needed to be present in order to use it :-) I did a first attempt 2 years ago, but I got lost in the composer API. I'll have a look at how you did it here.

MekDrop:
Ah... than see #629 It has some basic console command defined to install/uninstall module. The code is not perfect but after that I think it will be possible do the same with our composer plugin.

PR #616, opened 2020-03-21
Added module migrations support

MekDrop:
This pull requests adds support for module migrations from Modules Admin


fiammybe:
@MekDrop I will have to test this with an actual module. I propose https://github.com/IPFModules/billboard

MekDrop:
@fiammybe right now easier that convert into migrations from module that come with SQL file. Like iForum. If you would like, I could do that.

We still don't have a tool that could convert changes generated by object diffs into migrations like Billboard use. I have planned to rewrite imbuilding for that once all required changed where added in to core.

fiammybe:
:-) funny you mentioned iForum, I was looking into converting that one to a 'true' IPF module ASAP, given the issues we have on the website (I have SQL errors every time I want to post something).
But if you can use iForum as test case for this, that would be fine also. I'm not sure iForum is PHP7 compatible though.

MekDrop:
Ok than I will convert iforum. But I can't promise that I would do that sooner than in a week.

PR #618, opened 2020-03-22
Container defined editors and sourceeditors

MekDrop:
This pull request remakes how editors are detected and loaded. From now all editors and sourceeditors will be registered in container with tags `editor.source` and `editor.content` and new class that should implement EditorInterface.

What that means? From now is possible to register extra editors or sourceeditors from any module. So, I think, it will be easier for a user to replace one editor with another, because that can be done from same place as managing modules.


fiammybe:
Would it be possible to adapt the CKEditor 4 editor that we have, but that isn't included in the core at the moment, to show how it can take advantage of this improvement? 

I merged, did some non-regression tests and everything seems to continue to work, but I can't test the functionality you added in this PR without an example :-)

MekDrop:
To be able to use this feature, we need first make possible easy to fully install ImpressCMS modules with composer and than make a new module for some new editor (or move old editor there)

fiammybe:
OK, I'll leave this PR on the side then. moving on to the next!

MekDrop:
You know, I though a bit. And if this doesn't breaks anything I think it would be good to merge. Later it would be easier to implement my mentioned things.

PR #620, opened 2020-03-27
Routes defined in composer.json

MekDrop:
From now is possible to define routes in composer.json extra section.

Every route must have a `handler` and `path` defined. All handlers should be defined in `CONTROLLERCLASS::METHOD` format. It's callable that returns `\Psr\Http\Message\ResponseInterface`. Paths defines how route will be accessed from browser.

Example of such definition
```json
[
     {
        "path": "/error.php",
        "handler": "\\ImpressCMS\\Core\\Controllers\\DefaultController::getError"
      }
]
```

Every route can also have such defined options:
* method - HTTP method used for accessing this route (default GET)
* port - One or more ports list for this route
* host - One or more host list to be used with this route
* scheme - HTTP or HTTPS?
* strategy - json or app (default app). See https://route.thephpleague.com/4.x/strategies/
* middlewares - list of middleware classes (all middleware classes must be registered in container first) to be used with this route

For old modules there is already predefined LegacyController that is registered for most active modules files. 

Probably speed should be improved before release.


fiammybe:
I have been successful in installing this branch on top of the current master branch. However, once there, I cannot go anywhere because I get a 404 error : Error 404 Not Found
The page you are trying to access could not be found.

Are there standard routes that need to be defined into the current composer.json file in order to make it work?

MekDrop:
@fiammybe all routes real definitions should be found `storage/cache/routes.php`. There is a chance that something in other merged pull requestes made this file incorrectly generated. When I will come home I can look more into.

fiammybe:
@MekDrop okay, thanks. I'll have a look and let you know if I figure it out or not :-)

fiammybe:
The routes,php file is there, Could it perhaps be that I'm running in a subfolder?

MekDrop:
Yeah it seems the case. Installing in subfolder for unknown reason doesn't resolves to correct routes. Prefixing routes in cached routes path and editing .env file makes all routes partial accessible... so it seems it si solvable.

MekDrop:
I think now with subfolders everything will be fine. Currently there is small issue - when moving from folder to route URL (or from other side), routes cache must be cleared, but I think this is small thing so we could ignore that for a while.

fiammybe:
This now mostly works, as I have been able to update my testing setup. I say 'mostly', because in the ACP only the links to the adsense, autotasks, avatar, blocks, and comments work. Everything after that, starting with the custom tags, points in the end to the  index of the installation routine.
And you are right, it is really slow. Is some kind of caching possible?

MekDrop:
I started to look into what to do with speed but right now I don't have anything that could be usefull.

MekDrop:
So, removing shutdown function, adding few lines to properties improves speed very much. I think more improvements possible... but now it seems it's at least becomes useful.

fiammybe:
I don't see this kind of improvement, honestly. It is still extremely slow on my dev machine. When I look at the profiler data, 
League\Router\Route->dispatch takes up 12,9 seconds.
Drilling down, I can see that ImpressCMS\Core\Controllers\DefaultController->getIndex() is taking 40% of the processing time of the entire page (which is 20sec). Perhaps there is something that can be done there as well.

MekDrop:
@fiammybe can you try that with blackfire.io and share a link of the result with me? That would be helpful.

fiammybe:
OK, I now have 818ms as page generation time, and I strongly suspect that it is because of the xdebug that I disabled in order to get blackfire to run. I recall that composer had a similar problem some time ago. Speed-wise I think it is now ok. Remains the fact that most of the ACP links don't work at the moment. What generates the data in routes.php? Is it in the composer.json file?

MekDrop:
Old style links should be detected automatically when first time trying to load any page (except for error). New style links are defined in composer.json. And for me what I have checked randomly everything works so if deleting routes.php doesn't helps I think there are still some problems with old style routes logic.

PR #629, opened 2020-03-30
Added console commands support based on symfony/console

MekDrop:
This pull request adds normal console commands support and removes ` imponeer/composer-custom-commands` (which doesn't worked good enough to use for production) 

With this pull request comes few console commands:
* cache:clear - clears cache
* module:install - installs module
* module:uninstall - uninstalls module
* module:update - update module
* templates:cache:clear - clears compiled templates cache

All console commands should be registered in container with `console.command` tag. If command or any other container class implements `League\Container\ContainerAwareInterface` container will be binded automatically.


fiammybe:
I found this : /bin/console.bat (for use in windows) and it works. I don't know if there is a way of making it work from the root of the site?

PR #661, opened 2020-06-22
Fix empty login on Chrome-based browsers

### LINKED ISSUE #100: Login in Chrome points to blank page since 1.3.10 (and in 1.3.11beta)
This only happens on a site that is migrated from a previous version. 

After logging in, the user is presented with a blank page. Refreshing that page redirects to the correct user profile page.

fiammybe:
Fix #100
This will fix the long-standing bug that happens when you login from the user.php page (and not using the block on the homepage or somewhere else).

### DIFF
--- a/htdocs/user.php
+++ b/htdocs/user.php
@@ -50,7 +50,7 @@ $redirect = isset($_GET['xoops_redirect'])
 		? $_GET['xoops_redirect']
 		: isset($_POST['xoops_redirect'])
 			? $_POST['xoops_redirect']
-			: FALSE;
+			: ICMS_URL;
 if ($redirect) {
 	$redirect = htmlspecialchars(trim($redirect), ENT_QUOTES);
 	$isExternal = FALSE;


fiammybe:
Possibly. The reason we have a blank page is because, when you start from the user.php page to login, the redirect is empty, and in the end of the login process, ImpressCMS tries to redirect you to "", which won't work of course. 

I'll try to find out why in this case the problem is there and why it is not present when I am using exactly the same code in the same file when trying to login using the user block on the frontpage.

fiammybe:
In effect, you are correct. When I go to the /user.php page on a newly installed site, I have 2 ways of logging in : via the block in the side column, and via the block on the page itself. 

The block in the side column works as expected, and this is the form data from the request that is generated:
uname: xxxx
pass: yyyy
xoops_redirect: /test14/user.php
op: login

When I try to login in the main block of the page, the form data is missing content for the xoops_redirect variable:
uname: xxxx
pass: yyyy
op: login
xoops_redirect: 

Looking at the difference, in the block, the xoops_redirect parameter in the system_block_login.html template is using the $xoops_requesturi variable, where the user.php page is using the system_userform.html template with the $redirect_page variable.

When I put the $xoops_requesturi variable in the place of $redirect_page in the system_userform.html template, the login works as expected.

fiammybe:
correct implementation is on #670

PR #662, opened 2020-06-22
Fix the "Trying to access array offset on value of type null" 

fiammybe:
error in PHP 7.4 by checking if the variable is is set or not first

### DIFF
--- a/htdocs/install/page_siteinit.php
+++ b/htdocs/install/page_siteinit.php
@@ -86,7 +86,7 @@ ob_start();
 			<div class="dbconn_line"><label for="adminname"><?php echo ADMIN_DISPLAY_LABEL; ?></label>
 				<div class="clear">&nbsp;</div>
 				<input type="text" name="adminname" id="adminname" maxlength="25"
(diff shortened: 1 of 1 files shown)


htdocs/install/page_siteinit.php
```diff
@@ -57,60 +57,60 @@
...
-	value="" /></div>
-</div>
-</fieldset>
+	<fieldset>
+		<h3><?php echo LEGEND_ADMIN_ACCOUNT; ?></h3>
+		<div class="blokinit">
+			<div class="dbconn_line"><label for="adminname"><?php echo ADMIN_DISPLAY_LABEL; ?></label>
+				<div class="clear">&nbsp;</div>
+				<input type="text" name="adminname" id="adminname" maxlength="25"
+					   value="<?php if(isset($vars['adminname'])) { if(isset($vars['adminname'])) echo htmlspecialchars( $vars['adminname'], ENT_QUOTES );} ?>" />
```
MekDrop, line 89: 2 times same if?

fiammybe: too much copy-paste :-( Fixed now

PR #663, opened 2020-06-25
Making input boxes accept wider text strings

jegelstaff:
So email can be used comfortably as a username


fiammybe:
Thank you for the PR. I'll check if the new length gives problems in the database before accepting it.

fiammybe:
This is for 2.0?

jegelstaff:
No, it's just a fix. If you have a very long username, you can't enter it all in this box.

fiammybe:
The reason I ask is because the master branch is used for ImpressCMS 2.0. If you want it to appear in 1.4, could you please make a PR against the branches/impresscms_1.4 branch? Thanks!

PR #700, opened 2020-07-26
Fixes 'Deprecation Notice: Unparenthesized `a ? b : c ? d : e` is deprecated. Use either `(a ? b : c) ? d : e` or `a ? b : (c ? d : e)` in include/cp_functions.php:277'

MekDrop:
As title says


fiammybe (approved):
Less elegant, but if the previous way is deprecated, so be it :-)

PR #720, opened 2020-08-01
Using FireWall middleware for bad ips checking instead of Security class

### LINKED ISSUE #703: Use middlewares/firewall for dealing with IP restrictions
We have possibility to use accessing filtering by ip. We could use this middleware for better way dealing with that.

https://github.com/middlewares/firewall

MekDrop:
From now bad IPs checking uses https://github.com/middlewares/firewall that lets better to define single IPs or IPs ranges that should be banned. 

Also from security class removed `checkBadIp` method. 

Closes #703 

**Warning**: This pull request must be merged after #717 (it's based on this pull request code)


fiammybe:
Cool :-) I just saw that a recent mentioned Laravel package does the same thing more or less : https://github.com/Mahmoud-Italy/Larasecure-ipblocker 
As they say : great mind think alike :-)

fiammybe (approved):

PR #738, opened 2020-08-08
Use league/mime-type-detection for dealing with mimetype detection & deprecated icms_Utils

MekDrop:
### Resolves ImpressCMS/ideas#22 

From now system module doesn't have mimetypes editing section. This was some functionality that never worked as expected. Mimetype permission based on module maybe a good idea but current implementation was made a bit complex too what it really did and fully not worked as expected.

### Resolves ImpressCMS/ideas#31

`icms_Utils` is from now deprecated (marked to be removed in 2.1). [league/mime-type-detection](https://github.com/thephpleague/mime-type-detection) mime detection library should be used instead (same will be used in flysystem 2.0).


fiammybe (approved):
great to see some of the ideas that get implemented, and not just the existing but in a new technological way.

PR #742, opened 2020-08-20
Added Roave Security Advisories to composer [dev]

MekDrop:
Awesome [Php Inspections ​(EA Extended)](https://plugins.jetbrains.com/plugin/7622-php-inspections-ea-extended-) plugin has recommended that we should include [Roave Security Advisories](https://github.com/Roave/SecurityAdvisories) for keeping better project security.  So here it is.


fiammybe (approved):
Who are we to second-guess one of the best advisor modules for phpStorm :-) ?

fiammybe (approved):

PR #743, opened 2020-08-20
Run CI tests also for composer.json and composer.lock file changes

MekDrop:
I think composer files changes should be tested too! So, this pull request adds such tests.


fiammybe (approved):
right you are, the more we can test automatically the better

PR #753, opened 2020-08-23
Fixes bug when if value in criteria is not a string some comparisons fails

MekDrop:
(as title says)


fiammybe (approved):
improvement on a hack :-)

PR #762, opened 2020-09-02
Replace tecnickcom/tcpdf with dompdf/dompdf due license incompatibility

MekDrop:
[tecnickcom/tcpdf](https://github.com/tecnickcom/TCPDF) requires GPL 3.0 license but we use GPL 2.0 only. So I switch PDF rendering with [dompdf/dompdf](https://github.com/dompdf/dompdf). I hope this will be enough to solve this issue.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,6 +52,7 @@
     "aura/sqlschema": "^2.0",
     "cerdic/css-tidy": "^1.5",
     "composer/composer": "^1.10",
+    "dompdf/dompdf": "^0.8.5",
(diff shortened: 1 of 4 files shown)


fiammybe:
As a matter of fact, do we really NEED this PDF functionality in the core?

MekDrop:
I checked if have some modules that uses this functionality before starting this task and it seems we have few. So, I think atleast 2.1 this should be in the core.

fiammybe (approved):

PR #771, opened 2020-09-13
Fixed bug when ImpresCMS was installed in subfolder and route grouping functionality prevendted to add correct prefixes

MekDrop:
(as title says)

To make this work - clear cache first.


fiammybe (approved):
Works great!

PR #772, opened 2020-09-13
Fixes license in composer.json

MekDrop:
It seems we incorrectly said in our composer.json that we use GPLv2.0 only license but in reality our code is based on GPLv2.0 or later license (I have checked differences in our code base + all Xoops inherited file headers).


fiammybe (approved):
theis might reduce the number of issues FOSSA give us

PR #773, opened 2020-09-13
Fixed URLs for GPLv2 license in php files

MekDrop:
It seems license URL in many our PHP files pointed to incorrect location. Updated.

### DIFF
--- a/core/Extensions/Editors/EditorsRegistry.php
+++ b/core/Extensions/Editors/EditorsRegistry.php
@@ -30,11 +30,12 @@
 namespace ImpressCMS\Core\Extensions\Editors;
 
 use Exception;
+use icms;
 
 /**
  * Editor framework
  *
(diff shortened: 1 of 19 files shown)


fiammybe:
I guess PR #762 wasn't strictly necessary then :-)

fiammybe:
XOOPS has been very touchy about changing even a single letter in the original header. I am afraid we should leave the files that still have a link to XOOPS with the old link, even though it is wrong.

MekDrop:
See f.e. https://github.com/XOOPS/XoopsCore25/blame/master/htdocs/kernel/comment.php or any other file that we still using. They already did changes 5 years ago, so believe it's OK do same thing. Otherwise we need to upgrade into GPLv3-or-later and this could rise other problems for us to solve our license problems.

fiammybe (approved):

PR #788, opened 2020-10-26
Remove greetings actions due actions/first-interaction#23 bug

MekDrop:
At least until actions/first-interaction#23 is not resolved, removing this GitHub Action.


fiammybe (approved):
The errors were a pain, so better this way

PR #796, opened 2020-11-07
Using Composer 2.x API for internal operations

MekDrop:
Upgraded to use composer 2.x api for reading packages info internally.


fiammybe (approved):
simple change, not much I can add to that :-)

PR #800, opened 2020-11-08
Remove all editors from core

MekDrop:
In 2.0 after and in alpha 10 is possible to install editors as modules and this is only correct way that should be used in the future. This will helps us to easier keep track when editors source code are updated and also for any our user for freely to install selected editor.

There are not yet any module that installs editor for right but I hope to create at least one in next few weeks.

This removal also fix composer 2.0 issues.

Also, I recommend to review and merge #735 pull request too. Because it fixes some issues when DHTMLEditor is removed.


fiammybe:
I think I'll leave this one on hold until there is an editor module that we can test :-)

fiammybe:
as mentioned, don't merge until we have a new editor as module that is tested and is reliable.

skenow:
Stating the obvious - we must have an editor included, whether as a module or embedded.

Issue #805, opened 2020-11-15
Missing templates in system module


skenow:
## Describe the bug
Using the core image manager and the dhtmleditor, no images are displayed in the popup

## To Reproduce
1. Edit a text area using the dhtmleditor - in either the user side or admin area
2. Click on the imagemanager icon in the toolbar
3. See error

## Expected behavior
Image categories should be displayed and images can be viewed and selected

fiammybe:
I was under the impression that this was already fixed several months ago in the 1.4.1 release, to be honest.

skenow:
There were 5 or 6 still not in the list

PR #806, opened 2020-11-15
Include the missing templates for the image manager

skenow:
Fix for #805


skenow:
The only file I really changed was modules/system/icms_version.php - the others would not resolve because of formatting, EOL line differences, or something else I couldn't get past.

MekDrop (approved):

MekDrop:
@skenow I think you have enabled some autoformating feature in your editor so that's why here were so many changes. I think this doesn't breaks anything (or if 'yes', at least I didn't see what) and something improves, so I think maybe it's ok. 

Thanks!

PR #809, opened 2020-11-18
switch to a more explicit form of comparison

fiammybe:
to avoid cases where == would result in true, even when it isn't strictly the case

### DIFF
--- a/core/Security/Password.php
+++ b/core/Security/Password.php
@@ -92,7 +92,7 @@ final class Password {
 	 */
 	public function passExpired($uname = '')
 	{
-		if (!isset($uname) || (isset($uname) && $uname == '')) {
+		if (!isset($uname) || (isset($uname) && $uname === '')) {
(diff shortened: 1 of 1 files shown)


core/Security/Password.php
```diff
@@ -142,7 +142,7 @@ private function _passExpired($uname) {
 	 */
 	public function getUserSalt($uname = '')
 	{
-		if (!isset($uname) || (isset($uname) && $uname == '')) {
```
MekDrop, line 145: I think for this place, it would be better to use `empty($uname)` function. Less code similar result.

fiammybe: you mean this ? : if (!isset($uname) || (isset($uname) && empty($uname) ))

MekDrop: I mean you can replace all `!isset($uname) || (isset($uname) && $uname == '')` with one `empty($uname)` https://stackoverflow.com/a/15607428

MekDrop (approved):

PR #810, opened 2020-11-18
more strict comparison of variables

fiammybe:
same as #809 but for icms 1.4

### DIFF
--- a/htdocs/libraries/icms/core/Password.php
+++ b/htdocs/libraries/icms/core/Password.php
@@ -95,7 +95,7 @@ final class icms_core_Password {
 	 * @return   bool     returns true if password is expired, false if password is not expired.
 	 */
 	public function passExpired($uname = '') {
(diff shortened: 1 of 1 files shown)


htdocs/libraries/icms/core/Password.php
```diff
@@ -95,7 +95,7 @@ public function createCryptoKey($slength = 64) {
 	 * @return   bool     returns true if password is expired, false if password is not expired.
 	 */
 	public function passExpired($uname = '') {
-		if (!isset($uname) || (isset($uname) && $uname == '')) {
+		if (!isset($uname) || (isset($uname) && $uname === '')) {
```
MekDrop, line 98: use `empty` instead

MekDrop: plz, update this too!

MekDrop (approved):

PR #823, opened 2020-12-03
Temp fix: use composer v1.0 for tests

MekDrop:
Because we still can't use composer 2.0 due mindplay-dk/composer-locator#8 issue at least for a while we still require to use composer 1.0 for 2.0 installations.


MekDrop:
#826 has fixes for our main composer 2.x problem but it still doesn't has any fixes for editors. So, if we don't want to merge #800 we should merge this pull request first.

fiammybe (approved):

PR #826, opened 2020-12-04
Upgraded middlewares/referrer-spam to 2.0.2 for PHP 8.0 and Composer 2.0 supporr

MekDrop:
[middlewares/referrer-spam](https://github.com/middlewares/referrer-spam/issues/4#issuecomment-738474115) released new version of lib that solves old this lib compatibility problems with new technologies. So here is update!


fiammybe:
do you think it's a good idea to merge this one already, taking in to account the problems we have with composer 2? Or is this transparent for composer 2 users, and will this give us the ability to test on PHP8?

MekDrop:
composer 2.0 plugins are mostly backward compatible with composer 1.0 so there is no risk

fiammybe (changes requested):
shouldn't there also be a change in the composer.json file?

MekDrop:
I think updated composer.lock is only really needed.

MekDrop:
on another thought adding to composer.json min version constrain would help when we are merging barnches and having some conflicts... so I have fixed that...

fiammybe (approved):

PR #833, opened 2020-12-04
Remove the old FCKEditor - no longer supported

fiammybe:
This very old editor needs to be replaced.


MekDrop (changes requested):
It change requires also modifying user settings so some upgrade script for this also is needed to change user selection into another editor (without that it can break installation).

skenow:
Removing it will not affect a new install - the user choice of editors has not been made. It should also not affect an upgrade, unless we're deleting the files from the webroot, which we're not. The user's choice of editor will still be valid, as FCKeditor will still be present from their earlier install.

MekDrop:
@skenow I agree with your points but because this is also for security bug fix, I think old editor should be removed when upgrading otherwise this will not fix security issues.

skenow:
@mekdrop - if the patch removes FCKeditor, it first needs to install another editor if one does not exist. Then, it has to change the core admin settings and any module settings. It would be great if we could do this, but I believe it is beyond the scope of this and will only delay progress in other areas.

MekDrop:
yeah... maybe that would be too hard... but what I think at least we need to add warning in admin like that about writable mainfile.php if FCKEditor is still selected - if we can't fix automatically we should at least let to known about that for our users.

skenow:
I agree with that, @MekDrop. I know we're going to be able to manage editors like modules in 2.0 - so a warning like we have for the mainfile being writable, the install folder being present, and old versions of PHP and not using PDO could be added in this release with a link to instructions to remove things not being used.

fiammybe:
Will the presence of the vulnerable files still be a security problem?
Op 6/02/2021 17:58:52, Steve Kenow ***@***.***> schreef:
Removing it will not affect a new install - the user choice of editors has not been made. It should also not affect an upgrade, unless we're deleting the files from the webroot, which we're not. The user's choice of editor will still be valid, as FCKeditor will still be present from their earlier install.

skenow:
Yes, @fiammybe - the vulnerability is still present unless the files are modified or removed, even if the editor is not enabled for the site

skenow:
I have added a warning to the admin area - it will check if FCKeditor is selected as the default editor and recommend changing it and removing the folder. If it is not the default and still present on the server, the recommendation is just to remove it.

skenow:
I have another addition to make - during the upgrade, checking for any config option that matches FCKeditor is simpler than I thought. If the test is negative, we can attempt to remove the directory. The warning will still be there in case they need to manually remove, like we do with removing the install folder.

skenow:
Ready for review - upgrade checks to see if FCKeditor is selected (for all modules) and will remove the folders if it is not selected. Warnings have been added to the control panel for both its use and its presence. Language constants added and all languages have been brought up to date for global.php

PR #836, opened 2020-12-10
Limit maximum length of password

fiammybe:
Limit the length of the password to 99 positions. If the password is longer (and it can be sent using a direct request, the verification algorithm can take a long time to parse the value, leading to a potential DOS attack.

Issue identified as #1033373 by f1v3 on Cloudflare


htdocs/include/checklogin.php
```diff
@@ -48,7 +48,7 @@
 
 icms_loadLanguageFile('core', 'user');
 $uname = !isset($_POST['uname']) ? '' : trim($_POST['uname']);
-$pass = !isset($_POST['pass']) ? '' : trim($_POST['pass']);
+$pass = !isset($_POST['pass']) ? '' : substr(trim($_POST['pass']), 0, 99);
```
MekDrop, line 51: what if somebody already has pass with length > 99 ? :/

MekDrop (changes requested):
I think for this we will need to make some extra changes. F.e.: this should work also in password changing area and new user registration too. Also, not in only backend but also in frontend (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/password#maxlength )

htdocs/include/checklogin.php
fiammybe: if someone has a pass with length >99, I'll need to add a text that says you need to reset your password. Due to the hashing, we cannot know the length of the original password unfortunately

fiammybe:
I noticed in the frontend that the password length fields are almost all - except the password reset one, that was sloppy :-). So we can safely assume that those that are using the system normally will not have a password length over 32. I will update the allowed length from 99 to 32 then.

MekDrop (approved):

PR #837, opened 2020-12-16
Remove whitesource config

MekDrop:
Whitesource tries to scan .js files in composer packages installations and I didn't find any info how to prevent that, so I have decided to remove it.

We use Dependabot and HackerOne. These services I think could replace the need for Whitesource.


fiammybe (approved):
As a temporary measure, I agree. However I would look into reactivating whitesource in the future. It can be an indication for external libraries that might need to be updated

PR #838, opened 2020-12-16
Remove whitesource config

MekDrop:
Same as #837 but for 1.4.x


fiammybe (approved):
As a temporary measure, I agree. However I would look into reactivating whitesource in the future. It can be an indication for external libraries that might need to be updated

PR #844, opened 2020-12-17
Fixed null response bug for root path instalations

MekDrop:
Accedently I found that such bug exists. For such installations user will get fatal error on any page.


fiammybe (approved):
I guess this was a 'if' without handling the case where 'if' was false?

PR #851, opened 2020-12-21
Code Climate steps are now not required in GitLabCI test action

MekDrop:
We do not really need that code climate steps would succeed to make sure that everything ok - these steps just uploads tests results to codeclimate. So, I made them as optional.


fiammybe (approved):
great idea. Codeclimate is having some issues lately, so this will make sure our builds don't fail for that reason.

PR #852, opened 2020-12-21
anti-clickjacking

fiammybe:
Anti-clickjacking measure by adding a header("X-Frame-Options: SAMEORIGIN"); header directive


MekDrop (approved):
Good idea.

Still mozilla docs says thats this header should not be used in the future -> https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options But I think for a while that will be ok.

PR #854, opened 2020-12-30
Removed reflex theme from core

MekDrop:
Reflex theme is only theme included in our 2.0 distribution that doesn't work and idk how much it would be to fix that - this theme uses some variables that doesn't exists in current core. So, I have decided to remove it.


fiammybe (approved):

fiammybe:
First PR of 2021!

PR #866, opened 2021-01-01
Added filter-by-commitish option in release drafter config

MekDrop:
I'm not sure 100% but from documentation it seems this could fix issue when 1.4.x and 2.x are losing changes when regenerating new changelog.


fiammybe (approved):
I'm not sure release drafter is foreseen to work with a repository with multiple concurrent versions, but we can try this one

PR #869, opened 2021-01-02
Added ping to extend sessions automatically

MekDrop:
At current moment ImpressCMS doesn't automatically extends session if you leave your computer or browser tab for some time. This will helps users for such case. This could also help when you writing some big documents.  From now sessions are extended automatically by periodically pinging server with ajax requests.

In future I think it would be cool it could work with websockets too - but at current moment I think it would be good enough for us.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -570,6 +570,11 @@
         "handler": "\\ImpressCMS\\Core\\Controllers\\DefaultController::getError"
       },
       {
+        "path": "/ping",
+        "handler": "\\ImpressCMS\\Core\\Controllers\\DefaultController::getPing",
+        "method": ["GET"]
+      },
+      {
         "path": "/",
         "handler": "\\ImpressCMS\\Core\\Controllers\\DefaultController::getIndex",
         "method": ["GET", "POST"]
--- a/core/Controllers/DefaultController.php
+++ b/core/Controllers/DefaultController.php
@@ -2,10 +2,14 @@
 
 namespace ImpressCMS\Core\Controllers;
 
+use Aura\Session\Session;
+use GuzzleHttp\Psr7\Response;
+use icms;
 use ImpressCMS\Core\Response\RedirectResponse;
 use ImpressCMS\Core\Response\ViewResponse;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use function icms_loadLanguageFile;
 
 /**
  * Controller for doing some basic stuff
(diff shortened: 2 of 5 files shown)


skenow:
Does the session expire at some point? From a security standpoint, it must.

MekDrop:
@skenow when user close tab/browser, leaves website or internet connect is lost with the server. 

Most of popular websites with registrations uses pinging (similar to this) or websockets connection with server to know when user is not on the website anymore.  Also, this helps periodically regerate session id to make sure that for other users is harder to login as same user.

It's not ideal way but I think at current web standards it's a better than we have right now.

skenow:
@MekDrop we also have a session expiration setting in the admin control area, which 1st gets set when the user logs in and should be updated every time they visit the site. On our site, the setting is 30 minutes.

In my job, and I'm sure @fiammybe deals with it in his work, too - I deal with financial and personal information security and leaving browser sessions open is definitely not a good practice. I know most of the social platforms won't log you out for days, or ever, I'd like to have some control over it.

fiammybe:
@skenow the reasoning is sound here : it is possible, like I had yesterday, to be working on the site without doing a request to the server for an extended period of time. Our site 30 minutes is based on the idea that you will be doing a request to the server at least once in that timeframe. I must admit that it is increasingly not the case. 
It is true that at work, we develop mobile and web applications for all kinds of governement agencies, and there we implement something similar in Angular : user sessions get extended as long as the browser tab exists, with a maximum of extensions based on the preference of the particular DPO we're working with. Some want 4 hours, other 8. I think this is a parameter that should be configurable in the ACP.

MekDrop:
I could add option to turn on/off this feature (I'm not sure where). However something more I would like to posphone as much as posible - right now all new features are made almost by me and we still have many things to do :|

MekDrop:
I just thought: In ICMS 2.0 it would be possible to add such functionality with module. If such module would be included in default installation (selectable in install), I think maybe everyone would be happy and this functionality improvement would be not related to core. What do you think?

skenow:
In my mind, It's more the responsibility of the editor to maintain the session, not the core. People are frustrated when they are editing and composing, as you were, and their session expires before they save. Expiring at other times isn't as much of a problem - nothing is lost. CKEditor is configured with the Auto Save plugin as our default. The default is to destroy the local cache when the editor instance is destroyed - when you get logged out. That can be changed.

MekDrop:
@skenow problem not is only with editor but with forms overall. If user is adding some data with huge form session can also expire and user will get similar problem. 

Also from the security point autosaving editor at least in CKEditor 4 saves data in localstorage where is much easier to retrieve (you can open developer tools and see what is stored in local storage - you even don't need to have same user session) it for shared devices/computers than storing on the server. So, it's not a big win. 

CKEditor 5 uses similar way to store information as pinging. To be honest I would like to implement something like this but right now we don't have content versions support so I'm not sure how to implement such functionally.

skenow:
Should we just leave it, then, and have the administrator adjusts the session expiration to suit their needs?

skenow:
How about a popup message telling the user their session is about to expire. Banks and government websites here use that method.

MekDrop:
Leaving as is doesn't solve problem because if user wants to that he doesn't get timeouts he needs to increase session times on server. And sesssion times that are very long are much less secure than this - because no session id changes. Also, this way has a problem with server resources usage.

Yes popup is a good way for banks and goverment websites. It would be good have such support in core. But it how to implement that would work on every theme and look good? Right now we still don't have anything like that. Also, this way has a problem with users retention. I can't remember where I read but users hate automatic popups. So, that means site owner will loose visitors. So, I think as a variant for session expiring problem this could be included but would not make a default choice.

In any case ImpressCMS for banks or/and goverment websites is unusable in current state. We still doesn't have build-in support for even for GDPR requirements (personal data is not encrypted; no possibility to have data audit; no popup about external cookies and page how external services can be used) ;( Websites for banks and goverment has even more restrictions (like possibility to store user data per regions)!

fiammybe:
Let's add an option to turn it on/off and include it in the core. Saying it is the user's responsibility to make sure he does an action every 20 minutes is not userfriendly at all, and we need to be more user friendly if we ever want to regain more users (which is the main goal of icms 2.0). 
I encountered the problem when I was writing a long text, but I also get the problem when I am reading through long forum threads. So it is at the core level that we should provide a solution, and not just focusing on the use case of editing a text.

MekDrop:
Option tu turn on/off this feature was added

fiammybe (approved):

PR #882, opened 2021-01-10
Fixed some warnings and notices in installer for newer PHP versions

MekDrop:
Here was few small issues fixed that are visible when installing in newer PHP

### DIFF
--- a/htdocs/include/common.php
+++ b/htdocs/include/common.php
@@ -39,10 +39,11 @@
 defined("XOOPS_MAINFILE_INCLUDED") or die();
 
 /** @todo when this is no longer possible to run under PHP 5.x, we can remove the check - fiammybe 12 may 2019 */
-if(get_magic_quotes_runtime())
-{
-	// Deactivate
+/** @noinspection ConstantCanBeUsedInspection */
+if (version_compare(PHP_VERSION, '5.3.0', '<')) {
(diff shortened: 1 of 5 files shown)


htdocs/include/common.php
```diff
@@ -39,10 +39,11 @@
 defined("XOOPS_MAINFILE_INCLUDED") or die();
 
 /** @todo when this is no longer possible to run under PHP 5.x, we can remove the check - fiammybe 12 may 2019 */
-if(get_magic_quotes_runtime())
-{
-	// Deactivate
+/** @noinspection ConstantCanBeUsedInspection */
+if (version_compare(PHP_VERSION, '5.3.0', '<')) {
```
fiammybe, line 43: could you also use the PHP_VERSION_ID here, like in the other places?

fiammybe (commented):
just 1 question

htdocs/include/common.php
MekDrop: Yeah if you somewhere add at boot time these lines:
```php
if (!defined('PHP_VERSION_ID')) {
	$version = explode('.', PHP_VERSION);

	define('PHP_VERSION_ID', ($version[0] * 10000 + $version[1] * 100 + $version[2]));
}

```

fiammybe (approved):

PR #888, opened 2021-01-10
Added exception handler

MekDrop:
This will help easier to figure out what a blank page really means.

Example screen:
image

P.S.: syntax exceptions are handled >= 7.0; 5.6 handles most others; If exception happens before logger initialization also would not be handled in such way.


fiammybe:
Will this also catch the error messages that show the full path on the server?

MekDrop:
I can't say that will work for all but for some - yes

fiammybe (approved):

PR #889, opened 2021-01-10
Fix "#881 trying to send mails with SMTP auth gives missing smtp class"

### LINKED ISSUE #881: Trying to send mails with SMTPAuth gives mission 'SMTP' class
## Describe the bug
I want to send mails using the Google SMTP servers, because I need to have DKIM support activated to combat spam an phising from my domain. The Google SMTP servers use SMTPAuth. When I activate that, I get the following error message : 

Fatal error: Class 'SMTP' not found in /xxx/libraries/phpmailer/class.phpmailer.php on line 2604

## To Reproduce
1. Login to your site
2. Go the the ACP
3. Go to System > Site Configuration > Preferences > Mail Setup
4. Mail Delivery : SMTPAuth, SMTP Hosts : smtp.gmail.com, SMTPAuth username and password corresponding to your user, SMTP Secure method : SSL, SMTP Port : 465
5. Go to System > Users and Groups > Mail Users
6. identify some users, set a text and a subject, and send
7. Get error

## Expected behavior
I would like to system to use the GMail servers to send my mail.

## ImpressCMS
- version : 1.4.2

## Server
 - PHP version: 5.6
 - PHP is 32/64bit: 64
 - HTTP Server : [Apache, nginx] Litespeed (apache compatible)
 - HTTP Server version: ?
 - HTTP Server is 32/64 bit: 64

MekDrop:
Fixes #881 

Still for using google mail that could be noth enough - Google is moving to OAUTH and we do not have yet such functionality.

### DIFF
--- a/htdocs/libraries/icms/messaging/EmailHandler.php
+++ b/htdocs/libraries/icms/messaging/EmailHandler.php
@@ -43,6 +43,7 @@
  * load the base class
  */
 require_once ICMS_LIBRARIES_PATH . '/phpmailer/class.phpmailer.php';
+require_once ICMS_LIBRARIES_PATH . '/phpmailer/class.smtp.php';
 
 /**
  * Mailer Class.


fiammybe (approved):

fiammybe:
At the moment, you need to do an action to allow 'non-secure access' to the server. In the future, probably the access will need to be reviewed, yes.

PR #896, opened 2021-01-17
Fixed syntax error in include/registerform.php

MekDrop:
image
This will be fixed with after this pull request


fiammybe (changes requested):
best put everything back on 1 line

fiammybe (approved):

PR #919, opened 2021-02-07
Most of Smarty plugins now implemented as composer libraries from @imponeer + xoops_link smarty function removed

MekDrop:
This should be us easier to deal with Smarty plugins. 

Now each theme could define what from advanced Smarty syntax is used. So we can make sure what is really needed. 

Also, this will help us to deal better with licensing issues - allmost all smarty plugins were rewritten to be able use them with MIT license (and that is better for making comercial products).


fiammybe (approved):
I'm happy to see that this also reduces the complexity in the composer.json file :-)

PR #927, opened 2021-02-26
Use criteria lib from Imponeer

MekDrop:
Imponeer now [has rewritten database criteria classes with MIT license](https://github.com/imponeer/criteria). This pull request uses that.

It's one step closer to be able to write ICMS modules that doesn't require to include ImpressCMS into requirements.


fiammybe (commented):
Will this mean that every module will have to change the way they use criteria?

MekDrop:
@fiammybe in long term yes but at the moment old class aliases are also defined - so all modules should work in same way as previously.

fiammybe (approved):

PR #937, opened 2021-03-07
Fixes/ipf table filtering - limitsel missing POST value

### LINKED ISSUE #936: The limit selector does not change the number of items shown in ipf tables
## Describe the bug
In the admin area, many object are broken into several pages and the length of the page is determined by the limitsel field (defaults to 15). Changing the selection does refresh the page, but the number is not changed.

## To Reproduce
1. Go to Admin area, Blocks administration
2. Change the 'show only' field to another number (20)
3. After the page refreshes, the number shown is still 15

## Expected behavior
The number will update to your selected number and the number of items displayed on each page is updated

skenow:
The input filtering was not taking into consideration the POST changes. Fixes #936

### DIFF
--- a/htdocs/libraries/icms/ipf/view/Table.php
+++ b/htdocs/libraries/icms/ipf/view/Table.php
@@ -4,12 +4,11 @@
  *
  * Contains the classes responsible for displaying a highly configurable and features rich listing of IcmseristableObject objects
  *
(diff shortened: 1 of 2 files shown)


skenow:
The register.php was not in my commit - not sure why Git thinks it should be part of this PR

MekDrop (approved):

PR #943, opened 2021-03-14
Increase input sanitizing for system module and submodules

skenow:
Focused on standardizing input filtering using ICMS core filtering for all functions in the admin area.


fiammybe (approved):
It's impressive the number of locations where the handling of strings could be improved. Great work!

PR #967, opened 2021-05-26
filter url variable in findusers.php

fiammybe:
#fix ImpressCMS/security#18


MekDrop (approved):

htdocs/include/findusers.php
```diff
@@ -1,526 +1,526 @@
...
+						break;
+					case XOOPS_MATCH_CONTAIN:
+						$criteria->add(new icms_db_criteria_Item($var, '%'.$value.'%', 'LIKE'));
+						break;
+				}
+			}
+		}
+
+		if (!empty($_POST['url'])) {
+			$url = formatURL(trim(filter_input(INPUT_POST, $_POST['url'], FILTER_SANITIZE_URL)));
```
MekDrop, line 209: Incorrect ussage for filter_input. From what I understand, should be 
```php
filter_input(INPUT_POST, 'url', FILTER_SANITIZE_URL)
```

htdocs/include/findusers.php
```diff
@@ -1,526 +1,526 @@
...
+			$criteria->add(new icms_db_criteria_Item('url', $url.'%', 'LIKE'));
```
MekDrop, line 210: If url contains `%` character, criteria will return incorrect results. So, every `%` in url needs `\`. Source -> https://www.techonthenet.com/mysql/like.php

MekDrop (changes requested):
I found two issues with your work

htdocs/include/findusers.php
fiammybe: Thanks for the review. Do you mean the formatURL and the trim aren't necessary in this case?

MekDrop: I'm not sure about those, but I'm here about filter_input 2nd parameter

htdocs/include/findusers.php
fiammybe: ok, so that does mean that I need to transform every '%' to '\%' ?

htdocs/include/findusers.php
fiammybe: so I already need to assign $url = $_POST['url'] before then?

htdocs/include/findusers.php
fiammybe: @MekDrop hehe, github removed my escape character.I need to transform every percentage to baskslash-percentage by doing a replace?

htdocs/include/findusers.php
skenow: The second parameter is the $_POST variable - no need to pass $_POST['variable'], just the variable name as a string. Same with INPUT_GET.

htdocs/include/findusers.php
fiammybe: @MekDrop do you think it is best to define a new icms_core_Datafilter::addSlashesDB function, that escapes for the %, or do I extend the existing icms_core_Datafilter::addSlashes with the extra parameter that so that it mimics the addslashes PHP function?

MekDrop: I think probably adding extra parameter would be easier. But overall I don't see any big differences in these ways - we just need to make it work.

htdocs/include/findusers.php
fiammybe: done

MekDrop (approved):

PR #1017, opened 2021-10-13
Use editor contracts (second part)

MekDrop:
Because #1007 was too quickly merged this is will add other mixxing parts that should be merged.

It uses code from PR #1016 #1015 #1014 #800 code. So these must be merged first.


fiammybe:
sorry for the merge of the WIP :-(

PR #1077, opened 2021-11-02
Replaced league router with sunrise http router

MekDrop:
Sadly [League Router](https://route.thephpleague.com) for today doesn't have any possibilities to use reverse routing (aka  generate routes from route names). 

So after some experiments (my first choice was symfony router but it doesn't yet is compatible with PSR standards - so that's means many code lines needed to be rewritten) I have switched to [Sunrise HTTP Router](https://github.com/sunrise-php/http-router). It supports PHP 8.0 attributes and doctrine annotation style routes. Custom route sources also can be provided. So, I think it good enough for us. That's why I also removed support for routes definition in composer.json. I think now it will be easier to define routes.

Few other things:
 * [Flysytem](https://github.com/thephpleague/flysystem) was updated to 2.0
 * New `router:route-list` console command to debug active routes
 * Router definition now is done in service provider; same for dynamic middlewares - I think now it will be easier to edit these things
 * New ResponseFactory service that we can get from container by `response_factory` name can be used to get system response (f.e. for specific error code)


core/Middlewares/DetectModuleMiddleware.php
```diff
@@ -4,12 +4,17 @@
 
 use icms;
 use icms_module_Object;
-use ImpressCMS\Core\Controllers\LegacyController;
+use ImpressCMS\Core\Models\Module;
 use League\Route\Route;
```
fiammybe, line 8: still using league route here, I thought this was replaced now.

fiammybe (changes requested):
league/route is no longer is the composer file, but there are multiple references to the class still in the code.

htdocs/index.php
```diff
@@ -1,14 +1,11 @@
 <?php
 
-use Aura\Session\SessionFactory;
-use GuzzleHttp\Psr7\ServerRequest;
-use ImpressCMS\Core\Controllers\DefaultController;
 use ImpressCMS\Core\Database\DatabaseConnection;
-use ImpressCMS\Core\Facades\Config;
-use League\Route\Http\Exception as HttpException;
 use League\Route\Router;
```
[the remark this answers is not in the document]
MekDrop: Good point! Fixed.

core/Middlewares/DetectModuleMiddleware.php
MekDrop: And here fixed too!

fiammybe (approved):

PR #1079, opened 2021-11-06
Rewrite admin.php as controller

MekDrop:
*Warning* this pull request is based on #1077 so that pull request must be reviewed and merged first.

admin.php now rewritten as AdminController. So, -1 file in root path and better code.


fiammybe (approved):
I see the same changes from 1077 coming back again. Hope this doesn't break...

PR #1108, opened 2021-12-15
Add a default parameter to addSlashes

fiammybe:
This takes care of places where not second parameter is defined for addSlashes (like the login - oops)

### DIFF
--- a/htdocs/libraries/icms/core/DataFilter.php
+++ b/htdocs/libraries/icms/core/DataFilter.php
@@ -164,7 +164,7 @@ class icms_core_DataFilter {
 	 * Add slashes to the text if magic_quotes_gpc is turned off.
 	 *
 	 */
-	static public function addSlashes(string $text, string $param) {
+	static public function addSlashes(string $text, string $param = null ) {
 		if (!get_magic_quotes_gpc()) {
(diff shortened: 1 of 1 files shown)


htdocs/libraries/icms/core/DataFilter.php
```diff
@@ -164,7 +164,7 @@ static public function htmlEntities($text) {
 	 * Add slashes to the text if magic_quotes_gpc is turned off.
 	 *
 	 */
-	static public function addSlashes(string $text, string $param) {
+	static public function addSlashes(string $text, string $param = null ) {
 		if (!get_magic_quotes_gpc()) {
 			$text = addslashes($text, $param);
```
MekDrop, line 169: https://www.php.net/manual/en/function.addslashes.php seems doesn't have 2nd parameter. so, maybe we can just  remove $param from addSlashes ?

fiammybe: I forgot the other change : instead of calling addslashes, I now call addcslashes, which has the possibility to define for which data specifically you want to add slashes. That would be needed in order to define slashes for % in mysql

MekDrop (approved):

PR #1115, opened 2021-12-31
Accept token for valid users only

skenow:
The token should be tied to the user session and not allow anonymous access to this form


fiammybe (approved):
looks straightforward, but it has big implications for security. Nice!

PR #1125, opened 2022-01-23
Remove "roave/security-advisories"

MekDrop:
It seems we can't use it anymore, so removing - it doesn't do much changes for as at all - we have dependabot for that.


fiammybe (approved):
this package consistently posed an issue for me when building on Windows. Glad it's gone :-)

PR #1240, opened 2022-07-27
Upgrade dompdf to 2.0.0

fiammybe:
security fix


MekDrop (approved):

MekDrop (changes requested):
I approved and than I remembered that also `composer.lock` is here needed.

fiammybe:
I merged in the change that should automatically update the composer.lock, and then made some cosmetic changes to trigger a retest. Would that be enough to get the composer.lock up-to-date?

fiammybe:
Looks like it :-), I just now saw the new commit.

MekDrop (approved):

PR #1260, opened 2022-08-02
CI now tries automatically to autofix composer.lock files

MekDrop:
If composer.json is edited but composer.lock is not updated, CI action will try automatically to update composer.lock file with latest changes.

P.S.: because of checkout action bug doesn't correctly works with branches that has '.' in it's names


fiammybe (approved):
This is a great addition that will save us lots of time and tinkering, especially for small version updates. Thanks!

PR #1274, opened 2022-09-02
Remove openid

fiammybe:
Remove all occurences of OpenID in the core.


fiammybe:
I put it into draft because this is a huge PR, and It will need to be tested thoroughly before we not break 1.5 by merging it in :-)

MekDrop (approved):

PR #1283, opened 2022-09-27
Revert filesystems $provides array

MekDrop:
It seems without it now ICMS fails when displaying any page


fiammybe (approved):
I shouldn't have touched that array I think. Error on my side, sorry.

PR #1327, opened 2022-10-11
Remove get_magic_quotes_gpc calls

MekDrop:
This is for making it compatible with newer PHP versions.


fiammybe (approved):
Something screwed up the diff of the MediaUploadHandler.php, but the rest looks fine.

MekDrop:
@fiammybe it's because of line endings. If I remember correctly there was only lines 235, 241-242 changed

PR #1328, opened 2022-10-12
Fix: FILTER_SANITIZE_STRING is deprecated

### LINKED ISSUE #1159: FILTER_SANITIZE_STRING is deprecated in PHP 8.1
## Describe the bug
the use of FILTER_SANITIZE_STRING is no longer allowed. This creates problems in \core\DataFilter.php

MekDrop:
I think this is enough to fix #1159 issue


fiammybe (approved):
Bringing us on the road to PHP 8.1

PR #1342, opened 2022-11-21
replace create_function with anonymous function

fiammybe:
No longer available in PHP 8.0


htdocs/libraries/icms/view/block/Handler.php
```diff
@@ -219,7 +219,7 @@ public function getAllByGroupModule($groupid, $module_id = '0-0', $toponlyblock
 		$sql = "SELECT DISTINCT gperm_itemid FROM " . $this->db->prefix('group_permission')
 			. " WHERE gperm_name = 'block_read' AND gperm_modid = '1'";
 		if (is_array($groupid)) {
-			$gid = array_map(create_function('$a', '$r = "\'" . intval($a) . "\'"; return($r);'), $groupid);
+			$gid = array_map(function($a){$r = "\'" . intval($a) . "\'"; return($r);}, $groupid);
```
MekDrop, line 222: I think it could be also written here easier: ``$gid = array_map('intval', $groupid);`` Extra quotes here don't do much.

MekDrop (approved):

PR #1380, opened 2023-01-08
Fix XSS via DB_CHARSET parameter (H1 #1825770)

V35HR4J:
Fixed XSS via DB_CHARSET parameter while on installation.

Before:
Reported on https://hackerone.com/reports/1825770
Screen Shot 2023-01-08 at 11 30 33 AM

After:
sanitized error with htmlentities
Screen Shot 2023-01-08 at 11 30 56 AM

PR #1381, opened 2023-01-08
Fix XSS via DB_CHARSET parameter (H1 #1825770)

MekDrop:
Created from #1380 but for 1.5.x branch


fiammybe:
hehe, I was looking at the source as well, but you're faster than I am :-)

fiammybe (approved):

PR #1390, opened 2023-01-14
Fixed #1090 - user creating and editing

### LINKED ISSUE #1090: MySQL 8 compatibility
## Current code is not MySQL 8 compatible
Instances of database updates failing on MySQL 8

## Areas discovered
Adding new users
Registration of new users

## Server
 - MySQL 8.0.x

## Additional Context
When changing the MySQL version on the same server to 5.7, the above actions did succeed.

MekDrop:
Fix #1090 

It seems the problem was not with MySQL 8 but with the query code that the builder updated and insert user queries. Not always there were enough parameters for previously used `sprintf`.


fiammybe (approved):
Good find!

PR #1397, opened 2023-01-29
Fix ternary expression for php8

fiammybe:


htdocs/modules/system/icms_version.php
```diff
@@ -94,7 +94,7 @@
 /**
  * Manual
  */
-$modversion['manual']['wiki'][] = "<a href='https://www.impresscms.org/modules/simplywiki/index.php?page=Administration" . _LANGCODE . "' target='_blank'>" . _LANGNAME . "</a>";
+$modversion['manual']['wiki'][] = "<a href='https://www.impresscms.org/modules/simplywiki/index.php?page=Administration' target='_blank'>Manual</a>";
```
skenow, line 98: You can still use the language constants if you make sure to include the language files that contain these in install/makdedata.php, where @MekDrop made the other change

skenow (changes requested):

htdocs/modules/system/icms_version.php
fiammybe: Reverted this change.

skenow (approved):

PR #1398, opened 2023-01-29
upgrade cssTidy to 2.0.3, with support for PHP 8

fiammybe:


skenow:
From the readme - PHP 7.1 is the minimal version for this plugin. We'll have to address this in our requirements and upgrade path

fiammybe:
The base requirement for this release would be at least PHP 7.3, so the requirements for this plugin would be covered.

skenow (approved):

PR #1409, opened 2023-02-05
Update 1.5.x mimunum requirements

skenow:
Setting minimum PHP version to 7.0 based on PHP Code Sniffer and PHP Compliance tests. Some 3rd party libraries list PHP7.1 as their minimum, and still pass with 7.0. We can start with this and update as we move towards final release.


fiammybe (approved):
I made sure the version requirements are also updated in Dutch, Italian and French.

PR #1415, opened 2023-02-11
Errors in the Nederlands install language files

skenow:
Adjusting some quotes in response to errors on Scrutinizer


fiammybe (approved):
Clearly 2 cases of the cat running over my keyboard at those moments...

PR #1429, opened 2023-03-13
Move analytics code to preload

fiammybe:
Move the UA code to the preload. This code also looks for the value of the Google Analytics tag and decides based on that whether to add the old UA or the new GA4 script

### DIFF
--- a/htdocs/plugins/preloads/googleanalytics.php
+++ b/htdocs/plugins/preloads/googleanalytics.php
@@ -2,63 +2,41 @@
 /**
  * googleanalytics ga4.php
  * Created by david on 13/02/2022 22:41
- *
+ * Modified by skenow on 11 June 2023
  */
 
 /**
  * preload class to include the Google Analytics V4 tag in the Head of the webpage, after replacing the smarty variable with the GA4 property code
  * This can certainly be improved by checking whether we can, based on the preferences by our users (GDPR and cookies)
  */
-class IcmsPreloadGoogleanalytics extends icms_preload_Item {
+class IcmsPreloadGa4 extends IcmsPreloadItem {
 
-    function eventBeforeFooter()
-    {
-        global $xoopsTpl;
-        global $icmsTheme;
-        global $icmsConfigMetaFooter;
+	function eventBeforeFooter() {
+		global $xoopsTpl;
+		global $icmsTheme;
+		global $icmsConfigMetaFooter;
+		global $xoTheme;
 
+		try {
 
-        try {
-            if ($icmsConfigMetaFooter['use_google_analytics'])
-			{
-				if(isset($icmsConfigMetaFooter['google_analytics']) && $icmsConfigMetaFooter['google_analytics'] != '') 
-				{
-					if(substr( $icmsConfigMetaFooter['use_google_analytics'], 0, 2 ) === "G-")
-					{
-						$this->insertGA4Tag($icmsConfigMetaFooter['google_analytics']);
-					}
-					else
-					{
-						$this->insertUniversalAnalyticsTag($icmsConfigMetaFooter['google_analytics']);
-					}
+			if ($icmsConfigMetaFooter['use_google_analytics'] == TRUE && isset($icmsConfigMetaFooter['google_analytics']) && $icmsConfigMetaFooter['google_analytics'] != '') {
+				if (substr($icmsConfigMetaFooter['google_analytics'], 0, 2 ) === "G-") {
+					$tagvalue = $icmsConfigMetaFooter['google_analytics'];
+				} else {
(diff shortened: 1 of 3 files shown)


skenow:
In my tests, the ['async'] attribute is not rendered with this - I had to use ['async' => 'async'], which is valid notation for the script. Our code also inserts type=text/javascript into the script signature. Not a big deal.

fiammybe:
updated the 'async' => 'async' and making it ready for review

htdocs/plugins/preloads/googleanalytics.php
```diff
@@ -0,0 +1,64 @@
...
+		$icmsTheme->addScript('', '', '(function(i,s,o,g,r,a,m){i[\'GoogleAnalyticsObject\']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,\'script\',\'//www.google-analytics.com/analytics.js\',\'ga\');
+
+  ga(\'create\', \'UA-' . $UAtag  . '\', \'auto\');
+  ga(\'send\', \'pageview\');', 'module', '2001');
+
+	}
+
```
skenow, line 53: The current snippet from Google is a bit different and is almost identical to the GA4 snippet, using the Google Tag Manager. Should we switch?

skenow (changes requested):
Verify the proper tag values are being passed to the methods. The UA snippet has also changed when checking on GA.

htdocs/plugins/preloads/googleanalytics.php
```diff
@@ -0,0 +1,64 @@
...
+					$this->insertGA4Tag($icmsConfigMetaFooter['use_google_analytics']);
+				}
+				else
+				{
+					$this->insertUniversalAnalyticsTag($icmsConfigMetaFooter['use_google_analytics']);
+				}
+            }
+            else {
+                echo 'error getting icmsConfigMetaFooter';
+            }
```
skenow, line 35: Where will this display? The site may not have GA enabled or configured intentionally

htdocs/plugins/preloads/googleanalytics.php
skenow: It is identical - except for the tag value.

skenow: I've tested and the new snippet does work with older properties.

htdocs/plugins/preloads/googleanalytics.php
```diff
@@ -0,0 +1,64 @@
...
+        global $icmsConfigMetaFooter;
+
+
+        try {
+
+            if ($icmsConfigMetaFooter['use_google_analytics'] && isset($icmsConfigMetaFooter['google_analytics']) && $icmsConfigMetaFooter['google_analytics'] != '') {
+				if(substr( $icmsConfigMetaFooter['use_google_analytics'], 0, 2 ) === "G-")
+				{
+					$this->insertGA4Tag($icmsConfigMetaFooter['use_google_analytics']);
+				}
```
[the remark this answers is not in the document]
fiammybe: correct (2x) :-)

htdocs/plugins/preloads/googleanalytics.php
fiammybe: @skenow any word on this?

skenow: I'm not sure why I'm not being notified of comments here. Sorry to have missed this.

Having 1 function instead of 2 is cleaner, and when I look at the installation instructions for the UA tag in GA, it shows the same code as it does for the GA-4 tag. 

I think we should change it, so if admins don't switch to GA-4 on their own, when Google automatically does it, things should still work.

htdocs/plugins/preloads/googleanalytics.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        global $xoopsTpl;
+        global $icmsTheme;
+        global $icmsConfigMetaFooter;
+
+
+        try {
+
+            if ($icmsConfigMetaFooter['use_google_analytics'] && isset($icmsConfigMetaFooter['google_analytics']) && $icmsConfigMetaFooter['google_analytics'] != '') {
+				if(substr( $icmsConfigMetaFooter['use_google_analytics'], 0, 2 ) === "G-")
```
skenow: The parameter here should also be [google_analytics]

fiammybe: changed to the preload from the 1.4.6 branch.

skenow (approved):

PR #1430, opened 2023-03-20
1.5.x php7+

skenow:
Updates to ensure functionality on PHP7+ without errors


MekDrop (approved):
For me only one thing that comes to sigh is refs to 1.5 and sometimes 2.0. But maybe that in next PR could be fixed.

skenow:
I will reidentify the version in a coming PR

PR #1453, opened 2023-05-28
Updating PHPMailer for PHP8 support

skenow:
Updating to PHPMailer 6.8.0 for PHP8 support.


htdocs/libraries/icms/messaging/EmailHandler.php
```diff
@@ -1,157 +1,165 @@
...
  *
- * @category	ICMS
- * @package		Core
- * @subpackage	Mail
+ * @category ICMS
+ * @package Core
+ * @subpackage Mail
  */
-class icms_messaging_EmailHandler extends PHPMailer {
+class icms_messaging_EmailHandler extends PHPMailer\PHPMailer\PHPMailer {
```
MekDrop, line 57: can you use `use PHPMailer\PHPMailer\PHPMailer; on top of the file` and keep `class icms_messaging_EmailHandler extends PHPMailer {` here? I think, It would look something better :|

skenow: Thank you - I did make that change and tested on my server and it worked. I will adjust and commit.

MekDrop (approved):

PR #1464, opened 2023-06-17
Return correct types for custom session handlers

skenow:
session_set_save_handler called several methods that were described properly, but returned different types (object instead of boolean, for example). This caused fatal errors in PHP8

### DIFF
--- a/htdocs/libraries/icms/core/Session.php
+++ b/htdocs/libraries/icms/core/Session.php
@@ -61,7 +61,8 @@ class icms_core_Session {
 	static public function service() {
 		global $icmsConfig;
 		$instance = new icms_core_Session(icms::$xoopsDB);
-		session_set_save_handler(array($instance, 'open'), array($instance, 'close'), array($instance, 'read'), array($instance, 'write'), array($instance, 'destroy'), array($instance, 'gc'));
+		session_set_save_handler(array($instance, 'open'), array($instance, 'close'), array($instance, 'read'),
+			array($instance, 'write'), array($instance, 'destroy'), array($instance, 'gc'));
 		$sslpost_name = isset($_POST[$icmsConfig['sslpost_name']]) ? $_POST[$icmsConfig['sslpost_name']] : "";
 		$instance->sessionStart($sslpost_name);
 
@@ -186,7 +187,7 @@ class icms_core_Session {
 	 * @return bool
 	 */
 	public function write($sess_id, $sess_data) {
-		return self::writeSession($sess_id, $sess_data) ? true : false;
+		return (bool) self::writeSession($sess_id, $sess_data);
 	}
 
 	/**
@@ -196,7 +197,7 @@ class icms_core_Session {
 	 * @return bool
 	 */
 	public function destroy($sess_id) {
-		return self::destroySession($sess_id) ? true : false;
+		return (bool) self::destroySession($sess_id);
 	}
 
 	/**
@@ -206,7 +207,7 @@ class icms_core_Session {
 	 * @return bool
 	 */
 	public function gc($expire) {
-		return self::gcSession($expire) ? true : false;
+		return (bool) self::gcSession($expire);
 	}
 
 	/**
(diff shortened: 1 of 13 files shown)


htdocs/libraries/icms/core/Session.php
```diff
@@ -44,25 +45,23 @@
...
 	 * @return icms_core_Session
 	 */
 	static public function service() {
 		global $icmsConfig;
 		$instance = new icms_core_Session(icms::$xoopsDB);
-		session_set_save_handler(
-			array($instance, 'open'), array($instance, 'close'), array($instance, 'read'),
-			array($instance, 'write'), array($instance, 'destroy'), array($instance, 'gc')
-		);
+		session_set_save_handler(array($instance, 'open'), array($instance, 'close'), array($instance, 'read'), array($instance, 'write'), array($instance, 'destroy'), array($instance, 'gc'));
```
MekDrop, line 64: can you revert this to previous form? now it's harder to read - a bit too long

htdocs/libraries/icms/core/Session.php
```diff
@@ -89,144 +87,150 @@ static public function service() {
...
-	 * @return  bool
-	 **/
+	 *
+	 * @param string $sess_id
+	 * @param string $sess_data
+	 * @return bool
+	 */
 	public function write($sess_id, $sess_data) {
-		return self::writeSession($sess_id, $sess_data);
+		return self::writeSession($sess_id, $sess_data) ? true : false;
```
MekDrop, line 189: I think `(bool)self::writeSession($sess_id, $sess_data)` here could be a better - inline casting usually is faster than if

htdocs/libraries/icms/core/Session.php
```diff
@@ -89,144 +87,150 @@ static public function service() {
...
-	 * @param   string  $sess_id
-	 * @return  bool
-	 **/
+	 *
+	 * @param string $sess_id
+	 * @return bool
+	 */
 	public function destroy($sess_id) {
-		return self::destroySession($sess_id);
+		return self::destroySession($sess_id) ? true : false;
```
MekDrop, line 199: same here

htdocs/libraries/icms/core/Session.php
```diff
@@ -89,144 +87,150 @@ static public function service() {
...
-	 * @param   int $expire Time in seconds until a session expires
-	 * @return  bool
-	 **/
+	 *
+	 * @param int $expire Time in seconds until a session expires
+	 * @return bool
+	 */
 	public function gc($expire) {
-		return self::gcSession($expire);
+		return self::gcSession($expire) ? true : false;
```
MekDrop, line 209: same here

htdocs/libraries/icms/core/Session.php
```diff
@@ -446,43 +448,32 @@ private function readSession($sess_id) {
...
 		if (!icms::$xoopsDB->getAffectedRows()) {
-			$sql = sprintf(
-				"INSERT INTO %s (sess_id, sess_updated, sess_ip, sess_data)"
-				. " VALUES (%s, '%u', %s, %s)",
-				icms::$xoopsDB->prefix('session'),
-				$sess_id, time(),
-				icms::$xoopsDB->quoteString($_SERVER['REMOTE_ADDR']),
-				$sess_data
-			);
+			$sql = sprintf("INSERT INTO %s (sess_id, sess_updated, sess_ip, sess_data)" . " VALUES (%s, '%u', %s, %s)", icms::$xoopsDB->prefix('session'), $sess_id, time(), icms::$xoopsDB->quoteString($_SERVER['REMOTE_ADDR']), $sess_data);
```
MekDrop, line 463: I would prefer https://www.php-fig.org/psr/psr-12/#23-lines PSR-12 or some similar rule to keep up with line length - it's easier to read.

MekDrop (changes requested):

htdocs/libraries/icms/core/Session.php
skenow: Yeah, I don't like how the  formatter changed these lines either.

Eclipse does a fairly decent job and have configured the rules from the [ImpressCMS code style](https://www.impresscms.org/modules/simplywiki/index.php?page=Style_Guide) . How it handles arrays now is almost impossible to get to work as I want it to. And, in this file, arrays are arguments for another function, so it splits them even more.

htdocs/libraries/icms/core/Session.php
skenow: Changed it back - and working on the code formatter

htdocs/libraries/icms/core/Session.php
skenow: Applied and committed

htdocs/libraries/icms/core/Session.php
skenow: Applied and committed

htdocs/libraries/icms/core/Session.php
skenow: Applied and committed

MekDrop (approved):

PR #1465, opened 2023-06-18
Updating WideImage for PHP8 compliance

skenow:
Updating to version 1.1.4 - with support for PHP8. This removes the preload in favor of using namespaces and autoloading.


fiammybe (approved):
Pulled in the changes, and my version still works as before, which is the expected behaviour. Looks OK to me

PR #1469, opened 2023-07-01
Release 1.4.5 - 1.4 with GA4 fix

fiammybe:
Quick update to integrate the patch that @skenow  posted on the forums.


skenow (approved):
A release with this included is a good idea - thank you! I have this running on several sites and can confirm it is working.

PR #1476, opened 2023-07-23
get installer working on PHP 8.1

fiammybe:
These changes are needed to get the installer working on PHP 8.1. It also takes care of adapting the FILTER_SANITIZE_STRING which is deprecated.


htdocs/install/page_dbsettings.php
```diff
@@ -129,13 +130,13 @@ function getDbCharsets($link) {
 function getDbCollations($link, $charset) {
 	static $collations = array( );
 
-	if ($result = exec_query("SHOW COLLATION WHERE Charset=" . quote_sql($charset), $link)) {
+	if ($result = exec_query("SHOW COLLATION WHERE Charset=" . quote_sql($charset, $link), $link)) {
 		while ($row = fetch_assoc($result)) {
```
skenow, line 134: Should $link be here twice? quote_sql only takes one argument

skenow (commented):
I'm uncertain that the insertion of $link in several places of quote_sql is intended - the signature of that function only has 1 argument.

htdocs/install/page_dbsettings.php
fiammybe: that looks like a merge mistake, I'll verify to be certain.

fiammybe:
Now that I am looking at this request again, can you tell me again what the minimal requirements for 2.0 will be PHP-wise? If it is PHP 7.0 minimum (which is already VERY old), I can also remove all the checks for the old mysql routines, as it will then be PDO-only.

skenow (approved):
Those changes look good!

PR #1485, opened 2023-08-14
improve theme selector

fiammybe:
The theme selector was previously generating the HTML within the block routine, leaving very few options for themes to work with. This change updates that : the block routine in system_blocks.php now generates an array with the relevant information, and the system_block_themes.html template uses that info to create the view.


fiammybe:
Not the cleanest preparation on this one, sorry about that.

skenow (approved):
Checked out the changes and they work as desired

PR #1487, opened 2023-08-19
Fixes fatal error in googleanalytics preload

skenow:
The parent class is named incorrecty


skenow:
Up until 1.5, IcmsPreloadItem was still available from kernel/icmspreloadhandler.php and in the libraries/icms/Autoloader.php. That's why it worked there.

fiammybe (approved):
good catch! Strange this slipped in though, as I'm using a 2.0 RC version to test this.

PR #1496, opened 2023-08-24
language switcher refactoring

fiammybe:
This gives the themes access to the different variables of the language switcher, in order to decide on the layout. Not really backwards compatible with old themes

### DIFF
--- a/htdocs/modules/system/blocks/system_blocks.php
+++ b/htdocs/modules/system/blocks/system_blocks.php
@@ -460,13 +460,14 @@ function b_system_notification_show() {
  */
 function b_system_multilanguage_show()
 {
-	global $icmsConfigMultilang;
+	global $icmsConfigMultilang, $icmsConfig;
 
 	$block = array();
 
+
 	$language_tags = explode(',', $icmsConfigMultilang['ml_tags']);
 	$language_names = explode(',', $icmsConfigMultilang['ml_captions']);
-
+	$block['current_language'] = $icmsConfig['language'];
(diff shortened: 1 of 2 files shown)


skenow (approved):
This does work as intended. However, while testing I discovered that of the default values for the languages (English and German), the German language folder is not present, so only the label and link are present in the block (showing the English flag and the caption Deutsch). I suggest also changing the language constants _DEF_LANG_TAGS, _DEF_LANG_NAMES, _LOCAL_LANG_NAMES to have values of languages that are included in the core install

fiammybe:
Wasn't it the rule to have the languages in their native description? Instead of 'German' it should be 'Deutsch' then. Similar for 'Nederlands' (Dutch) or 'Italiano' (you guessed it, Italian).


On Sat, Sep 02, 2023 at 12:21am, Steve Kenow < ***@***.*** ***@***.*** > wrote:


@skenow approved this pull request.

This does work as intended. However, while testing I discovered that of the default values for the languages (English and German), the German language folder is not present, so only the label and link are present in the block (showing the English flag and the caption Deutsch). I suggest also changing the language constants _DEF_LANG_TAGS, _DEF_LANG_NAMES, _LOCAL_LANG_NAMES to have values of languages that are included in the core install

skenow:
I think you could call it anything you wanted, as long as everything matched - the language folder, the flag icon, the name you use in the multi-language settings. All of the select boxes scan the language folder for the folders present and present them as they are named in the folder. Using native descriptions would be ideal

PR #1500, opened 2023-09-03
Smiley adminstration input filtering

skenow:
Standard filtering was applied at the head of the file, later code segments did not.


skenow:
There may be a slight difference in how this works without an additional change I made in my local environment while troubleshooting - I altered icms_core_DataFilter::checkVar so it wouldn't return false when the data being checked was 'falsey', like a zero. I'll test more and actually create a new PR for the change in that method.

fiammybe (approved):

PR #1527, opened 2023-10-24
Making sure password reset key is unique and temporary

skenow:
Change in the structure of the data in the field over time has made the old partial string common for all users. This change makes it unique and temporary, basing a portion of the key on their last login. Cleaned up the email subjects a bit, too.

### DIFF
--- a/htdocs/lostpass.php
+++ b/htdocs/lostpass.php
@@ -59,11 +59,11 @@ $code = '';
 if (!empty($_GET)) {
 	// in places where strict mode is not used for checkVarArray, make sure filter_ vars are not overwritten
 	if (isset($_GET['filter_post'])) unset ($_GET['filter_post']);
-    $clean_GET = icms_core_DataFilter::checkVarArray($_GET, $filter_get, FALSE);
+    $clean_GET = icms_core_DataFilter::checkVarArray($_GET, $filter_get, false);
     extract($clean_GET);
 }
 if (!empty($_POST)) {
-    $clean_POST = icms_core_DataFilter::checkVarArray($_POST, $filter_post, FALSE);
+	$clean_POST = icms_core_DataFilter::checkVarArray($_POST, $filter_post, false);
     extract($clean_POST);
 }
 if ($email == '') {
@@ -82,7 +82,7 @@ if (empty($getuser)) {
 } else {
 	$icmspass = new icms_core_Password();
 
-	$areyou = substr($getuser[0]->getVar('pass'), 0, 5);
+	$areyou = substr($getuser[0]->getVar('pass'), -5) . $getuser[0]->getVar('last_login');
 	if ($code != '' && $areyou == $code) {
 		$newpass = $icmspass->createSalt(8);
 		$pass = $icmspass->encryptPass($newpass);
@@ -97,7 +97,7 @@ if (empty($getuser)) {
 		$xoopsMailer->setToUsers($getuser[0]);
 		$xoopsMailer->setFromEmail($icmsConfig['adminmail']);
 		$xoopsMailer->setFromName($icmsConfig['sitename']);
-		$xoopsMailer->setSubject(sprintf(_US_NEWPWDREQ, ICMS_URL));
+		$xoopsMailer->setSubject(sprintf(_US_NEWPWDREQ, $icmsConfig['sitename'] . ' (' . ICMS_URL . ')'));
 		if (!$xoopsMailer->send()) {
 			echo $xoopsMailer->getErrors();
 		}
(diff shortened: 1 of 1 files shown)


fiammybe (approved):

htdocs/lostpass.php
```diff
@@ -97,7 +97,7 @@
 		$xoopsMailer->setToUsers($getuser[0]);
 		$xoopsMailer->setFromEmail($icmsConfig['adminmail']);
 		$xoopsMailer->setFromName($icmsConfig['sitename']);
-		$xoopsMailer->setSubject(sprintf(_US_NEWPWDREQ, ICMS_URL));
+		$xoopsMailer->setSubject(sprintf(_US_NEWPWDREQ, $icmsConfig['sitename'] . ' (' . ICMS_URL . ')'));
```
MekDrop, line 100: I think it would be more correctly to add ICMS_URL into language file as new variable than doing that here. 

That constant comes from user.php -> https://github.com/ImpressCMS/impresscms/blob/eb9403215854d733d01a2f73c5370d81cd9ca2dc/htdocs/language/english/user.php#L89 

So change that line into: 
```php
define('_US_NEWPWDREQ','New Password Request at %s (%s)');
```

And here:
```php
$xoopsMailer->setSubject(sprintf(_US_NEWPWDREQ, $icmsConfig['sitename'], ICMS_URL));
```

This would make sure that everywhere same type of text is used/displayed in same way.

htdocs/lostpass.php
```diff
@@ -127,7 +127,7 @@
 		$xoopsMailer->setToUsers($getuser[0]);
 		$xoopsMailer->setFromEmail($icmsConfig['adminmail']);
 		$xoopsMailer->setFromName($icmsConfig['sitename']);
-		$xoopsMailer->setSubject(sprintf(_US_NEWPWDREQ, $icmsConfig['sitename']));
+		$xoopsMailer->setSubject(sprintf(_US_NEWPWDREQ, $icmsConfig['sitename'] . ' (' . ICMS_URL . ')'));
```
MekDrop, line 130: same here

htdocs/lostpass.php
```diff
@@ -82,7 +82,7 @@
 } else {
 	$icmspass = new icms_core_Password();
 
-	$areyou = substr($getuser[0]->getVar('pass'), 0, 5);
+	$areyou = substr($getuser[0]->getVar('pass'), -5) . $getuser[0]->getVar('last_login');
```
MekDrop, line 85: I think it would be more secure if `$areyou` would be hashed value

F.e.:
```php
$areyou = hash('crc32', substr($getuser[0]->getVar('pass'), -5) . $getuser[0]->getVar('last_login'));
```

This will output 8 char length code like '9064a8d0' and it will hide all components needed for that code to be generated. In current way is possible to guess that after trying many code reminders.

ChatGPT for the problem solution recommends to use md5 hash:
F.e.:
```php
$areyou = md5(substr($getuser[0]->getVar('pass'), -5) . $getuser[0]->getVar('last_login'));
```

So, code will look like '65a8e27d8879283831b664bd8b7f0ad4' and will be always 32 chars length.

First hash variant I think is easier for mobile users - if they are needed to enter code manually. The second is more secure.

MekDrop (changes requested):
I think at least language constants parts should be changed but you also should be think about the hashing. Take a look into my comments before.

skenow:
I don't think anyone will be manually entering the key - and there is no way for a user to do that except adding it to the URL. 

Hashing the key would keep from exposing actual data in the user's record, once they know the format (read the code). The key also changes the next time the user logs in, which is why I added that to the key. This was simpler than adding another field to the database for password reset request timestamp and count, where we can control how long the key lasts and how many requests can be made before another action is required.

Having a consistent subject line is what I was going for - in one message it had the site name, in the other it had the URL. Looking at the language files, there are similar constants - 
define('_US_YOURACCOUNT', 'Your account at %s');

htdocs/lostpass.php
fiammybe: If we hash, md5 is outdated and we should use sha1 now at the least. But I do agree hashing is the way to go.

skenow: I have added hashing to the key. I have also added a check so if the key doesn't match, it doesn't generate another email, keeping the user from being flooded if there is a brute force attack. The language constants - I've left them. They could be improved and the sprintf() can be removed. _US_RESETPWDREQ is defined and never used.

htdocs/lostpass.php
skenow: I propose this -
`define(_US_NEWPWREQ', New Password Request at . $icmsConfig['sitename'] . ' (' . ICMS_URL . ')'));`

With this, the sprintf() in lostpass.php can be eliminated.

htdocs/lostpass.php
MekDrop: For including icmsConfig and ICMS_URL in constant value I see only one possible problem - I think we a service that we use for easy translations - crowdin - doesn't support including variables directly into values. Idk how others do. Also that isn't compatible with i18n where we must one day to go (I think sprintf translations kinda compatible with thatmutch more)

fiammybe (approved):

PR #1553, opened 2024-06-25
Cleanup social provider list

fiammybe:
remove old providers


MekDrop (approved):

MekDrop:
Honestly, I'm not sure why social providers names are in constants defined... doesn't do much... but still cleanup is really nice thing!

fiammybe:
I thought the same, but rebuilding that block entirely would take much more time. I'd rather do a small improvement a day, instead of launching myself in big refactoring work at the moment. The goal is to get 2.0 over the finish line in the best condition possible, the soonest possible.

PR #1554, opened 2024-07-17
Update version.php to new 2.0.0 beta 2

fiammybe:
the versioning of 2.0.0 beta 1 was a bit off, correcting that for the beta 2 release beforehand

### DIFF
--- a/htdocs/include/version.php
+++ b/htdocs/include/version.php
@@ -8,7 +8,7 @@
  * @since Xoops
  * @author phppp
  */
-define('ICMS_VERSION_NAME', 'ImpressCMS 2.0.0 RC2');
+define('ICMS_VERSION_NAME', 'ImpressCMS 2.0.0 Beta 2');
 
 // For backward compatibility with XOOPS
 define('XOOPS_VERSION', ICMS_VERSION_NAME);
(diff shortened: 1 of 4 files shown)


htdocs/modules/system/icms_version.php
```diff
@@ -66,7 +66,7 @@
 /**  Development information */
 	'status_version' => "RC",
```
skenow, line 67: Should probably change this to beta, too

fiammybe: As a matter of fact, this would be RC2, instead of Beta 2. We're coming from RC1.

htdocs/include/version.php
```diff
@@ -8,7 +8,7 @@
  * @since Xoops
  * @author phppp
  */
-define('ICMS_VERSION_NAME', 'ImpressCMS 2.0.0 RC2');
+define('ICMS_VERSION_NAME', 'ImpressCMS 2.0.0 Beta 2');
```
skenow, line 12: RC, right?

htdocs/include/version.php
```diff
@@ -21,16 +21,16 @@
  * 10 = Final
  */
 
-define('ICMS_VERSION_STATUS', 3);
+define('ICMS_VERSION_STATUS', 2);
```
skenow, line 25: 2 = beta

htdocs/include/version.php
fiammybe: I was mistakenly convinced that it was RC already, but beta it is!

htdocs/include/version.php
fiammybe: As we're going for Beta, that's correct.

skenow (approved):
Everything seems to be in alignment now

PR #1559, opened 2024-08-03
Remove banner tables and config item

### LINKED ISSUE #1528: Upgrade does not remove banners db tables or configuration entries
## Description
banner, bannerclient, and bannerfinish all remain in the database after an upgrade; ACP > Preferences > General : Activate banner ads option still exist

## Additional context
The installation procedures and language files have been updated so new installs don't have these, we need to adjust for upgrades

skenow:
This will remove the database tables for banners and the config item  left from previous versions of the core Resolves #1528

### DIFF
--- a/htdocs/modules/system/include/update.php
+++ b/htdocs/modules/system/include/update.php
@@ -48,10 +48,10 @@ icms_loadLanguageFile('core', 'databaseupdater');
  * @param int $dbVersion The database version
  * @return mixed
  */
-function xoops_module_update_system(&$module, $oldversion = NULL, $dbVersion = NULL) {
+function xoops_module_update_system(&$module, $oldversion = null, $dbVersion = null) {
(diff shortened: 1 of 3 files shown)


htdocs/modules/system/include/update.php
```diff
@@ -105,17 +105,13 @@ function xoops_module_update_system(&$module, $oldversion = NULL, $dbVersion = N
 
 	/* Begin upgrade to version 1.5 */
 	if (!$abortUpdate) {
-		$newDbVersion = 47;
+		$newDbVersion = 48;
 	}
```
skenow, line 109: I changed this because I was running the beta version on my site. I didn't think to add another section. I can if you think it best

skenow: I have created separate blocks for beta 1 and beta 2 - this should be resolved.

MekDrop (approved):

PR #1563, opened 2024-08-22
cleanup icms.css

fiammybe:


skenow:
I overlooked removing it from being included.

Are we also removing icms.css and icms_rtl.css? What brought this to my attention was a style in icms.css was changing how my pages were displaying. Are there elements that need to move to the base theme?

skenow (approved):
This removes it from being included. Still need to remove the files and add any necessary elements to the theme's css

PR #1584, opened 2024-10-22
Templated form elements

fiammybe:
CSS frameworks (Bulma, Bootstrap, Tailwind, ...) and almost every theme out there need specific HTML markup for their form solution to work well. The HTML implementation in the core was not flexible enough to allow that. I started with a button, and will gradually transform all hardcoded HTML in the core to using templates. 

I've started with these 4:
* Text Area
* Password
* Button
* Checkbox

Other parts will come in other PRs 

Smarty is included in the core specifically for this kind of functionality, so let's not re-invent the wheel. 

Added benefit : by using Smarty we should also have template caching, for that extra micro-second of speed :-)

### DIFF
--- a/htdocs/libraries/icms/form/Element.php
+++ b/htdocs/libraries/icms/form/Element.php
@@ -116,6 +116,13 @@ abstract class icms_form_Element {
 	 * @var	string
 	 */
 	private $_description = "";
+
+	/**
+	 * template for this field
+	 */
+	protected icms_view_Tpl $_tpl;
+	protected string $_customTemplate = "";
+
 	/**#@-*/
 
 	/**
--- a/htdocs/libraries/icms/form/Element.php
+++ b/htdocs/libraries/icms/form/Element.php
@@ -120,8 +120,8 @@ abstract class icms_form_Element {
 	/**
 	 * template for this field
 	 */
-	protected icms_view_Tpl $_tpl;
-	protected string $_customTemplate = "";
+	protected icms_view_Tpl $tpl;
+	protected string $customTemplate = "";
 
 	/**#@-*/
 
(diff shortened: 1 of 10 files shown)


htdocs/libraries/icms/form/Element.php
```diff
@@ -116,6 +116,13 @@ abstract class icms_form_Element {
 	 * @var	string
 	 */
 	private $_description = "";
+
+	/**
+	 * template for this field
+	 */
+	protected icms_view_Tpl $_tpl;
```
MekDrop, line 123: can we for new vars not use `_` prefix? it's doesn't make sense

htdocs/libraries/icms/form/elements/Button.php
```diff
@@ -116,6 +116,16 @@ public function getType() {
 	 * @return	string
 	 */
 	public function render() {
-		return "<input type='".$this->getType()."' class='formButton' name='".$this->getName()."'  id='".$this->getName()."' value='".$this->getValue()."'".$this->getExtra()." />";
+	$this->_tpl = new icms_view_Tpl();
+	$this->_tpl->assign('type', $this->getType());
+	$this->_tpl->assign('name', $this->getName());
+	$this->_tpl->assign('id', $this->getName());
+	$this->_tpl->assign('value', $this->getValue());
+	$this->_tpl->assign('extra', $this->getExtra());
```
MekDrop, line 124: some formatting issue I think is here? somehow it's looks without indent

MekDrop (changes requested):

htdocs/libraries/icms/form/Element.php
fiammybe: I believe it was in the early PHP4 days a convention to show that it was a protected variable. You are right, it doesn't make any sense anymore and I'll change it.

MekDrop (approved):

PR #1588, opened 2024-11-02
Remove old columns from the users table Resolves #1561

### LINKED ISSUE #1561: OpenID columns still exist in the users table after an upgrade
users.openid and users.user_viewoid remain after an upgrade to 2.0. These columns should be removed.

skenow:
This will remove the openid and user_viewoid columns from the users table during an upgrade from pre-2.0 releases.


fiammybe (approved):
This works, and it can stay for this version, but I have a feeling this is a functionality that could benefit to be made available in a more general way in the future.

skenow:
I've been thinking about this - the system module would need a very large overhaul and a lot of table would need to be renamed for it to work as other IPF modules work. Definitely something to plan

PR #1595, opened 2024-11-25
Convert to current global for $icmsModule

### LINKED ISSUE #1593: Several areas affected by using $icmsModule instead of icms::$module
- In testing the custom CKeditor toolbars based on the user group, I discovered the permissions were not being followed because the dhtmltextarea form element handler was using the global $icmsModule, which is no longer available.
- Permission for viewing debug by a group for a module
- Several system blocks don't render or are empty

skenow:
Regression testing uncovered several core elements that needed adjustment after the change in the global for $icmsModule. This converts the remaining variable references and resolves #1593

### DIFF
--- a/htdocs/include/notification_select.php
+++ b/htdocs/include/notification_select.php
@@ -43,9 +43,9 @@
 if (!defined('ICMS_ROOT_PATH')) {
 	exit();
 }
-include_once ICMS_ROOT_PATH.'/include/notification_constants.php';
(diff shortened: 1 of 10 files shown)


fiammybe (approved):
Incredible I could have missed so many. I did do a search in the code, but I'm starting to suspect I had local changes already that weren't on Github. Some of these locations don't ring a bell, so you will certainly have found more instances than I did.

PR #1601, opened 2024-12-26
Fix:Fix the CKEditor Image Manager that couldn't find WideImage

### LINKED ISSUE #1599: Imagebrowser plugin for CKeditor error when trying to browse images
Edit an HTML area, click on the image selector, click on Browse Server, then click on the browse category (magnifying glass). Fatal error - unable to load WideImage

fiammybe:
fixes #1599


MekDrop (approved):

skenow (commented):
It took me a bit to understand this - and why it's necessary in the first place to have the ternary comparison. We could probably simplify a lot of places this is used by having the method properly returning a boolean value instead

PR #1616, opened 2025-01-12
fix:in some cases, the update to version 48 was not persisted in the DB

fiammybe:
This resulted in the question to upgrade the system module each time


MekDrop (approved):

skenow:
In the first section to update to dbversion 47, the call to update the dbversion is part of the try{} portion, the last thing before catch{}. In the section for dbversion 48, it is after catch{}. All of other upgrade files that use try-catch have the dbversion update as part of try{}

PR #1619, opened 2025-02-02
fix:the DB update to v48 is now correctly positioned

### LINKED ISSUE #1617: Please review upgrade 48
Changes made in #1616 placed the last dbversion update completely outside of the try-catch handling, different than all other version upgrades using this approach.

fiammybe:
the fix is now correctly placed inside the try/catch structure
fixes #1617


skenow (commented):
Compare this section with the section for $dbversion = 47. Should these both be in the if($dbVersion < $newDbVersion){} portion, or after it? Before implementing the try-catch methods, the dbupdate was in the if() portion, not afterwards.

I think this one is correct, the one for 47 will need to align with this. See also update-14.php

fiammybe:
@skenow the merge should be clean now.

skenow (approved):
Looks good

PR #1623, opened 2025-03-22
Fix: icms_core_Object cleanVars and getVar errors with empty arrays

### LINKED ISSUE #1621: icms_core_Object::getVar returns false for an array type when the string cannot be unserialized
For data type XOBJ_DTYPE_ARRAY, getVar expects a string that can be unserialized and should return an array if the unserialization succeeds. However, when it fails, it should return an empty array. Instead, it returns 'FALSE'. Further functions and methods that expect an array and get a boolean instead throw a bunch of errors.

skenow:
cleanVars was skipping empty arrays and not serializing them causing errors. 
getVar was not returning an array if the string could not be unserialized, also causing errors.

Close #1621


htdocs/libraries/icms/core/Object.php
```diff
@@ -437,7 +437,7 @@ public function getVar($key, $format = 's') {
 				break;
 
 			case XOBJ_DTYPE_ARRAY:
-				$ret = unserialize($ret);
+				$ret = unserialize($ret) ? unserialize($ret) : array();
```
MekDrop, line 440: `unserialize` sometimes can throw exceptions, so I think this is safer approach:
```suggestion
	try {
        $ret = unserialize($data);
        if ($ret === false) {
            $ret = [];
        }
    } catch (\Throwable $e) {
        $ret = [];
    }
```

MekDrop (approved):

PR #1627, opened 2025-05-03
Search for user URL fixed

### LINKED ISSUE #1620: User profile URLs won't save without the protocol (http or https)
I believe this is a result of filtering we've done to improve security and in order for the URL to be determined as valid, it needs a protocol.
A side effect of this is searching users for URL contents does not return users whose URL does contain the protocol. To see an example of this, go to the administration area of ImpressCMS.org, go to Find Users, and first search on all users and sort by number of posts descending. Next, search again, this time with a 'w' in the URL contains field, sort by the same (# of posts, descending). My profile will show up in the first search and not the second, even though my URL contains a w (4 of them, to be accurate)

skenow:
Close #1620 

Input for search by URL was searching for URLs starting with, not contains. There are a few other slight changes for formatting and default protocol when saving URLs (changed to https instead of http)


htdocs/include/functions.php
```diff
@@ -171,7 +171,7 @@ function formatURL($url)
 	$url = trim($url);
 	if($url != '')
 	{
-		if((!preg_match("/^http[s]*:\/\//i", $url)) && (!preg_match("/^ftp*:\/\//i", $url)) && (!preg_match("/^ed2k*:\/\//i", $url))) {$url = 'http://'.$url;}
+		if((!preg_match("/^http[s]*:\/\//i", $url)) && (!preg_match("/^ftp*:\/\//i", $url)) && (!preg_match("/^ed2k*:\/\//i", $url))) {$url = 'https://'.$url;}
```
MekDrop, line 174: this small change would make it correctly work also with [about](https://en.wikipedia.org/wiki/About_URI_scheme), [acct](https://en.wikipedia.org/wiki/Acct_URI_scheme), [blob](https://en.wikipedia.org/wiki/Blob_URI_scheme), [data](https://en.wikipedia.org/wiki/Data_URI_scheme), [file](https://en.wikipedia.org/wiki/File_URI_scheme), [ftpes](https://en.wikipedia.org/wiki/FTPS), [ftps](https://en.wikipedia.org/wiki/FTPS), [geo](https://en.wikipedia.org/wiki/Geo_URI_scheme), [intent](https://en.wikipedia.org/wiki/Android_software_development), [irc](https://en.wikipedia.org/wiki/IRC_URI_scheme), [itms](https://en.wikipedia.org/wiki/ITunes_Store), [jdbc](https://en.wikipedia.org/wiki/JDBC), [msteams](https://en.wikipedia.org/wiki/Microsoft_Teams), [mailto](https://en.wikipedia.org/wiki/Mailto), [market](https://en.wikipedia.org/wiki/Google_Play), [psns](https://en.wikipedia.org/wiki/PlayStation_Network), [sip](https://en.wikipedia.org/wiki/SIP_URI_scheme), [sips](https://en.wikipedia.org/wiki/SIP_URI_scheme), [shortcuts](https://en.wikipedia.org/wiki/Shortcuts_(Apple)), [slack](https://en.wikipedia.org/wiki/Slack_(software)), [sms](https://en.wikipedia.org/wiki/SMS_URI_scheme), [ssh](https://en.wikipedia.org/wiki/Secure_Shell), [stratum](https://en.wikipedia.org/wiki/Stratum_protocol), [sftp](https://en.wikipedia.org/wiki/SSH_File_Transfer_Protocol), [steam](https://en.wikipedia.org/wiki/Steam_(software)), [tel](https://en.wikipedia.org/wiki/Tel_URI_scheme), [trueconf](https://en.wikipedia.org/wiki/TrueConf), [urn](https://en.wikipedia.org/wiki/Uniform_Resource_Name), [view-source](https://en.wikipedia.org/wiki/View-source_URI_scheme), [viber](https://en.wikipedia.org/wiki/Viber), [ws](https://en.wikipedia.org/wiki/WebSocket), [wss](https://en.wikipedia.org/wiki/WebSocket), [xmpp](https://en.wikipedia.org/wiki/XMPP), [zoommtg](https://en.wikipedia.org/wiki/Zoom_Video_Communications)...:
```suggestion
		if (!parse_url($url, PHP_URL_SCHEME)) {
		    $url = 'https://'.$url;
		}
```

htdocs/modules/system/admin/findusers/main.php
```diff
@@ -266,8 +275,8 @@
 		}
 	}
 	if (!empty($user_url)) {
-		$url = formatURL(trim($user_url));
-		$criteria->add(new icms_db_criteria_Item('url', $url . '%', 'LIKE'));
+		$url = filter_var($user_url, FILTER_SANITIZE_URL);
+		$criteria->add(new icms_db_criteria_Item('url', '%' . $url . '%', 'LIKE'));
```
MekDrop, line 279: I'm not 100% sure here but I think this is better then stripping special characters (it should escape them):
```suggestion
		$criteria->add(new icms_db_criteria_Item('url', '%' . str_replace(['%', '_'], ['\%', '\_'], $user_url) . '%', 'LIKE'));
```

MekDrop (approved):
kinda fixes the issue so I'm accepting PR but I think maybe there are some even better ways to improve the code... I left few suggestions... maybe at least one of them would be useful ;)

htdocs/modules/system/admin/findusers/main.php
skenow: Escaping them would make them safe to add to the database - what are the risks when the link is rendered and someone clicks on it?

htdocs/include/functions.php
skenow: I was curious about the choice of schemes and had to search to find ED2K is for eDonkey, something I still know little about. I was just looking at a more reasonable default.

htdocs/modules/system/admin/findusers/main.php
MekDrop: For outputing, I think logic is in getVar method, if correct mode is used it shouldn't be a problem. But in any case is much less risky then the issue with DB. It would be much harder to exploit that.

skenow: % and _ are valid characters in a URL and there are other characters that should be removed or encoded/escaped. 
Maybe this would be a better filter option? https://www.php.net/manual/en/filter.constants.php#constant.filter-sanitize-encoded

skenow: Using formatURL() is going to result in the same issue that started this - the user input is going to be converted to something with 'https://' prefixed at the beginning. What we need is an SQL-safe value from user input to query the database for a URL that contains the input, and it should only be for characters that are allowed in a valid URL

fiammybe:
Question here : do we need to allow a search via URL?

htdocs/modules/system/admin/findusers/main.php
MekDrop: Yeah... somehow I overlooked that... and AI from small context window not allways correctly returns response... :|

Probably this should be:
```suggestion
$criteria->add(
   new icms_db_criteria_Item(
      'url', 
      '%' . str_replace(['%', '_'], ['\%', '\_'], icms_core_DataFilter::addSlashes(trim($user_url))) . '%',
      'LIKE'
   )
);
```

Also, I noticed others criteria incorrectly escapes special LIKE characters

skenow: icms_core_DataFilter::checkVar() is already filtering the user input at the beginning of modules/system/admin/findusers/main.php and the default for any variable not specified in filter_post() or filter_get() is to treat as a string input. The default filtering for that is 

`return` htmlspecialchars(strip_tags($data), ENT_NOQUOTES);

Do we need more?

MekDrop: icms_core_dataFilter doesn't have anything that would escape correctly data for LIKE queries, so I'm thinking that extra escaping is needed. 

See this little test -> https://onlinephp.io/c/9ee76947-3320-4f64-94e1-63b110f5c054 

But still I'm not sure if there really is a big problem

skenow: > % and _ are magic symbols in MySQL LIKEs, so what here happening is these symbols are escaping.

I wasn't paying attention to that - it makes sense to escape those characters. I performed a test and searching for '%' returns all users with websites, not just ones with that character.

I still feel we should sanitize the input using FILTER_SANITIZE_URL - why let them search for something that should not be there?

skenow: Question (still learning) - would switching to prepared statements also be effective in escaping the input prior to executing the query?

MekDrop: Switching to prepared statements would also address the overall security issue. However, if I recall correctly, it took me over a month to implement that on the TNG branch. It would be great to backport that work - it would save a lot of time.

About another your question - I'm unsure. Usually if such questions arise I'm writing tests. But how do that for such code, I'm really unsure.

fiammybe:
I'll include this in the 2.0.2 release then. We can improve in later versions

PR #1629, opened 2025-05-04
Adding automatic update of the system module

skenow:
adsense, autotasks, customtag, mimetype, rating are all fully IPF-compliant and can benefit from automatic updates.


MekDrop (approved):

MekDrop:
I'm not sure if is correct to call this automatic updates. Somehow when I'm hearing this word I'm thinking what packages managers does (but this is another thing) but in any case your solution seems works... so approve ;)

fiammybe (approved):

PR #1651, opened 2025-09-01
fix:templated checkbox doesn't take checked state into account

fiammybe:
The template didn't take into account the 'checked' state. 

* After reviewing the code, I refactored it to have the information in a single array with sub-values, instead of 3 different arrays that need to be consolidated in the template. 
* I added the same logic to the IPF variants of the classes so all checkboxes, password boxes, buttons and text boxes now use a single template and should look the same, whether constructed using IPF or not.

### DIFF
--- a/htdocs/libraries/icms/form/elements/Checkbox.php
+++ b/htdocs/libraries/icms/form/elements/Checkbox.php
@@ -197,6 +197,9 @@ class icms_form_elements_Checkbox extends icms_form_Element {
 			$ele_name = $ele_name . "[]";
 			$this->setName($ele_name);
 		}
+		if (count($ele_value) > 0 && in_array($value, $ele_value)) {
+			$ele_ischecked = "'checked'";
+		}
 
 		$this->tpl = new icms_view_Tpl();
 		$this->tpl->assign('ele_name', $ele_name);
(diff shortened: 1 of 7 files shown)


skenow:
I've been trying to figure out how to download just the files that have been updated. It looks like I'll have to download the entire branch and then pick and choose so I can test. 

Question: It looks as though you have created a new folder in the root path for templates. How will we be using this folder and what order can things be overridden? Currently, the template is in the module's template folder and can be overridden by a theme by a template of the same name in the theme/module/modulename/templates/ folder

skenow:
Another question - the templates in /templates/ are the same as the ones added in /modules/system/templates/. Are they needed in both places?

fiammybe:
good catch @skenow , those were duplicates that somehow got included on the wrong location. I fixed that in the meantime.

skenow (commented):
This does resolve this issue. The added tests for ICMS_ROOT_PATH and the templates folder seem unnecessary, unless you've found otherwise.

fiammybe:
@skenow you are correct, I misread the path and didn't catch that it was looking for a template folder beneath de root. That will never happen, so I removed that piece of code.

PR #1655, opened 2025-09-21
Fix 1550 db collation prefill in mysql8+

### LINKED ISSUE #1550: fix:Database Collation prefills no longer there during installation on MySQL 8
Looks like the names of the database collation information has changed. We should now use utf8mb4 as character set (uft8 is no longer there), combined with utf8mb4_bin collation

fiammybe:
we now propose utf8umb4, as that is proposed to be the best 'universal' one.
close #1550


skenow (approved):
This will make the install a little smoother!

PR #1658, opened 2025-09-25
Fix empty page on table without operation

fiammybe:
It shouldn't be needed to always have an operation parameter in the request. The default should be to show the table.
Currently, without operation, you get a white page. This check validates that there is an operation present, and if not just renders the page (expected behaviour).

### DIFF
--- a/htdocs/install/page_dbsettings.php
+++ b/htdocs/install/page_dbsettings.php
@@ -91,11 +91,9 @@ function sanitize_database($database_name) {
 
 function quote_sql($sql) {
 	global $link;
-	if ($link instanceof PDO) {
+
 		return $link->quote($sql);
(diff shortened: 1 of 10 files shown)


skenow (commented):
Where are you seeing this and getting errors? When I go to adsense or custom tags after a fresh install, for example, there are no items in the table and the page renders properly.

This code block gets used in a lot of places, and only runs if there is something in the $_GET or  $_POST variable. Is there something that can be done to minimize impact everywhere it is used?

skenow:
imBlogging and imTagging also do not have any items to display initially and the admin page loads without error for those modules. There are no $_GET or $_POST values for them. Same with the submodules adsense and custom tags.

PR #1686, opened 2026-02-25
Keep-alive the session using a  preloader

fiammybe:
Nothing is as annoying as finishing a really long post, pushing the 'publish' button and being presented with a login page... and your post is gone. This preload sets up a 5-minute ping to the server for logged in users, so as long as your browser is open, you can extend your session.

3 files:
* keepalive.php in the root of the site - the scripts that gets called to extend the session
* plugins/preloads/keepalive.php - the script that installs the javascript ping on every page for logged-in users.
* assets/js/keepalive.js - the javascript that takes care of calling keepalive.php in the root of the site, to extend the session.

I've kept is as simple as possible (fixed 5 min delay, no configuration), but I wanted to have the functionality focused on real users that have a benefit. guest users can't edit normally, and shouldn't have these session extensions.


skenow:
What security considerations do we need? Several security reports did mention sessions as something to be careful with

fiammybe:
Good question. This PR checks for the user to be logged in before extending the session. Going over the code now I notice that there is no use of a CSRF token, so I'll put that in to harden even further.

fiammybe:
@skenow I simplified the javascript part, as I came to realize that the whole activity tracking wasn't really needed. The idea is to be plain and simple : as long as the browser is open for a logged-in user, the session gets extended every 5 minutes. I did this simplification after some important security hardenings as well. 
What we do is load the core, and let the core handle the entire session business. There is a basic flood protection in place, but it's basic.

htdocs/plugins/preloads/keepalive.php
```diff
@@ -0,0 +1,46 @@
...
+
+		/* ------------------------------------------------------------------
+		 * Register the external script, attaching a unique id
+		 * and the keep‑alive endpoint as a data attribute.
+		 */
+		$keepaliveUrl = ICMS_URL . "/keepalive.php";
+
+		$xoTheme->addScript(
+			"assets/js/keepalive.js",
+			[
```
skenow, line 32: I think you'll need to specify the path differently and use the full path

fiammybe: correct, I'll have to review that. Thanks!

PR #1701, opened 2026-07-26
Transform the IPF Classlibrary to PSR-4 and PHP 8.2

fiammybe:
Transform the existing classlibrary to PSR-4 structure and PHP 8.2+.
* 'Object' classes are renamed to 'Entity' due to a PHP 7.2 language restriction
* The 'icms' folder is renamed to Icms (with a capital 'I') to conform PSR4 specs and the fact that the class names start with a capital
* Pass by reference is removed in function definitions, following current PHP recommendations.
* Every PSR-4 class has an alias that points to the old classname. This should maintain complete backward-compatibility in modules that haven't yet converted to using namespaces.

This builds on the changes in https://github.com/fiammybe/impresscms/tree/ipf-to-composer


fiammybe:
This is turning out to be a huge PR. I will follow this up with PRs for smaller class categories, that will make it easier to do a follow-up.

fiammybe:
I'm doing a separate push of every part of the library, one folder at a time. Every part should also have associated tests to verify that the alias exists and the legacy code should keep working.

skenow:
This is a lot of work! I have been scrolling through the file differences and it all seems quite solid. As we move forward with this, there is a nagging consideration - file systems that are case sensitive will see all these as new folders and files and there will be a lot of cleanup to do.

fiammybe:
Thanks! It is quite mechanical work, but I want to verify everything manually and not let AI handle this all alone. 
I also thought about the case-sensitive aspect. The best final scenario would be to move the IPF classes to a separate composer library, which means removing them from the /libraries/icms folder into composer-managed folders. Until we are there, we could rename 'icms' to 'Icms' upon upgrade I guess?

skenow:
Renaming icms to Icms would be part of the process - also would need to rename all the folders and files beneath it. Maybe it would be easier to rename the existing folder to icms_legacy (or something similar) and add everything in a new Icms folder

Oh - and thank you for putting your expertise ahead of AI!

fiammybe:
going from icms to icms_legacy also means renaming folders, which is the tricky thing we're trying to avoid if possible. Perhaps keep all the original files as legacy, with only the class alias in them? And the PSR-4 transformed files would go into a composer library from the start, or a group of composer libraries, depending on how we structure this

skenow:
Maybe I'm not looking at it properly. In the list of changes for this PR, files and folders are being removed from /icms/ and added to /Icms/, which is the desired result. Where the challenge will be is on a non-case-sensitive operating system and being able to determine what you're dealing with during an upgrade.

fiammybe:
I believe that is how Git interprets a rename of a folder : remove and add again of the files in those folders, as Git only stores the content of files. A folder without a file in it can't be managed by Git.

PR #1711, opened 2026-08-31
Improve CSRF error handling in pmlite

fiammybe:
This pull request updates htdocs/pmlite.php to display specific security error messages directly in the browser when validation fails, instead of just setting an internal flag. The code now shows a warning message with a prompt and button to return to the previous page.
* Replaced the error variable assignment with a direct HTML output for security warnings.
* Added text, an action button, and a list of detailed errors.
* No change to the existing user count query logic that follows the changes.

### DIFF
--- a/htdocs/pmlite.php
+++ b/htdocs/pmlite.php
@@ -63,8 +63,13 @@ xoops_header();
 if (icms::$user) {
 	if (isset($_POST['op']) && $_POST['op'] == "submit") {
 		if (!icms::$security->check()) {
-			$security_error = true;
+			echo "<br /><br /><div><h4>" . _PM_USERNOEXIST . "<br />"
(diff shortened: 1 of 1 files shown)


htdocs/pmlite.php
```diff
@@ -63,8 +63,13 @@
 if (icms::$user) {
 	if (isset($_POST['op']) && $_POST['op'] == "submit") {
 		if (!icms::$security->check()) {
-			$security_error = true;
+			echo "<br /><br /><div><h4>" . _PM_USERNOEXIST . "<br />"
+				. _PM_PLZTRYAGAIN . "</h4><br />";
```
skenow, line 67: The test was for the token, not a user - right? The message content suggests the requested user does not exist.

